PR #2, opened 2016-12-21
Extensions - Collections

portellaa:
Add some trivia operations into Collections

* Array
* Dictionary

### DIFF
--- a/Alicerce/Sources/Extensions/Collection.swift
+++ b/Alicerce/Sources/Extensions/Collection.swift
@@ -1,0 +1,51 @@
+//
+//  Collection.swift
+//  Alicerce
+//
+//  Created by Luís Afonso on 18/12/2016.
+//  Copyright © 2016 Mindera. All rights reserved.
+//
+
+import Foundation
+
+// MARK: - Array
+
+func -<Key: Hashable, Value: AnyObject>(left: [[Key : Value]], right: [[Key : Value]]) -> [[Key : Value]] {
+    var final = left
+    
+    for (index, element) in right.enumerated() {
+        final[index] = (left[index] - element)
+    }
+    
+    return final
+}
+
+// MARK: - Dictionary
+
+func -<Key: Hashable, Value: AnyObject>(left: [Key : Value], right: [Key : Value]) -> [Key : Value] {
+    var final = left
+    
+    for (key, value) in right {
+        if let val = left[key], val === value {
+            final[key] = nil
+        }
+    }
+    
+    return final
+}
+
+func +<Key, Value>(left: [Key : Value], right: [Key : Value]) -> [Key : Value] {
+    var final = left
+    
+    for (key, value) in right {
+        final[key] = value
+    }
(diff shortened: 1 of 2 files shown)


Alicerce/Sources/Extensions/Collection.swift
```diff
@@ -0,0 +1,51 @@
...
+//  Collection.swift
+//  Alicerce
+//
+//  Created by Luís Afonso on 18/12/2016.
+//  Copyright © 2016 Mindera. All rights reserved.
+//
+
+import Foundation
+
+// MARK: - Array
```
p4checo: This should be `Dictionary Array` 😛

Alicerce/Sources/Extensions/Collection.swift
```diff
@@ -0,0 +1,51 @@
...
+    
+    return final
+}
+
+// MARK: - Dictionary
+
+func -<Key: Hashable, Value: AnyObject>(left: [Key : Value], right: [Key : Value]) -> [Key : Value] {
+    var final = left
+    
+    for (key, value) in right {
```
p4checo: Just to be clearer I would perhaps rename this `value` to `lValue` or `leftValue`, and the `val` below `rValue` or `rightValue` 😉

Alicerce/Sources/Extensions/Collection.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+    
+    return final
+}
+
+func +<Key, Value>(left: [Key : Value], right: [Key : Value]) -> [Key : Value] {
+    var final = left
+    
+    for (key, value) in right {
+        final[key] = value
```
p4checo: This behavior might not be obvious, because we overwrite existing values on `left` with the ones on `right` when `key` matches. Don't know if adding a "non destructive" version would be useful though 🤔

Alicerce/Sources/Extensions/Collection.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+    
+    return final
+}
+
+func +=<Key, Value>(left: inout [Key : Value], right: [Key : Value]) {
+    for (key, value) in right {
+        left[key] = value
```
p4checo: Same as above

Alicerce/Sources/Extensions/Collection.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+//  Created by Luís Afonso on 18/12/2016.
+//  Copyright © 2016 Mindera. All rights reserved.
+//
+
+import Foundation
+
+// MARK: - Array
+
+func -<Key: Hashable, Value: AnyObject>(left: [[Key : Value]], right: [[Key : Value]]) -> [[Key : Value]] {
```
p4checo: It would be nice to add some documentation to these operators 😇 

Personally, I prefer documenting with three slashes `///` instead of `/**/`.

PR #4, opened 2017-03-08
Service locator UTs

portellaa:


AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+    
+    var serviceLocator = ServiceLocator()
+    
+    override func setUp() {
+        super.setUp()
+        
+        serviceLocator = ServiceLocator()
+    }
+    
+    func testRegisterUsingServiceTypeInference() {
```
portellaa: yes, makes a lot of sense. that is the problem when we use testing frameworks 😄 

👍

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+            let rDoubleService: Double = try serviceLocator.get()
+            let rIntService: Int = try serviceLocator.get()
+            
+            // Validate registered services
+            XCTAssertEqual(rStringService, testServiceString)
+            XCTAssertEqual(rDoubleService, testServiceDouble)
+            XCTAssertEqual(rIntService, testServiceInt)
+            
+        } catch let error {
+            XCTAssert(false, "💥: Register failed with error: \(error)")
```
portellaa: thanks 😅

p4checo:
We can guide ourselves using the following checklist for our `ServiceLocator` UTs, considering the current public API's:

- `register` (normal)
	+ Success
		* name ✅ 
		* type ✅ 
	+ Failure
		* duplicate service
		* duplicate lazy service

- `register` (lazy)
	+ Success
		* name ✅ 
		* type ✅ 
	+ Failure
		* duplicate service
		* duplicate lazy service

- `unregister`
	+ Success
		* name ✅ 
		* type ✅ 
	+ Failure
		* inexistent service
		* service type mismatch
		* lazy service type mismatch (WIP)

- `unregisterAll`
	+ Success ✅ 

- `get`
	+ Success
		* name
		* type
	+ Failure
		* inexistent service
		* service type mismatch
		* lazy service type mismatch (WIP)

After these I think we can consider trickier tests like recursiveness and other wacky use cases we can think of 😆

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+            XCTAssertEqual(rStringService, testLazyServiceString())
+            
+        } catch let error {
+            XCTAssert(false, "💥: Register failed with error: \(error)")
+        }
+    }
+    
+    func testRegisterUsingServiceTypeInference_GetUsingName() {
+        
+        typealias DoubleService = Double
```
portellaa: fixed 👍

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+            XCTAssertEqual(rStringService, testServiceString)
+            XCTAssertEqual(rDoubleService, testServiceDouble)
+            
+            
+        } catch let error {
+            XCTAssert(false, "💥: Register failed with error: \(error)")
+        }
+    }
+    
+    func testRegisterUsingName_GetUsingTypeInference() {
```
portellaa: sure, i can remove them but if we return the name when we register, we can do things like this

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+            
+            // Check if it contains
+            XCTAssertEqual(rStringService, testServiceString)
+            
+            try serviceLocator.unregister(testServiceString)
+            
+            let _: String = try serviceLocator.get()
+            
+        } catch ServiceLocatorError.inexistentService {
+            XCTAssert(true, "🎉 Service removed with success")
```
portellaa: wut?

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+    
+    func testUnregisterAll() {
+
+        let testServiceString = "Test Register Service"
+        let testServiceDouble = 1.0
+        let testLazyServiceInt: () -> (Int) = {
+            return 1
+        }
+
+        func assert(with error: Error) {
```
portellaa: 🥇

AlicerceTests/ServiceLocatorTests.swift
p4checo: Yes, like we discussed that brings some value to these `register` API's and also makes testing way easier 😉

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,301 @@
...
+            } catch ServiceLocatorError.inexistentService {
+                XCTAssert(true, "🎉 Service \(String(describing: type)) removed with success")
+            } catch let error {
+                assert(with: error)
+            }
+        }
+
+        // Test that services were all removed
+
+        checkServiceRemoved(Int.self)
```
portellaa: Already discussed this in person. 💪 

OK, i agree with one dict for  the services but not the api for `list()` i think we don't need it.
Regarding the `enum`, we discussed we don't need the service `Type` for validation.

portellaa:
@p4checo ready for review.

portellaa:
I think i implemented all the tests, but will wait for your PR

AlicerceTests/ServiceLocatorTests.swift
p4checo: Agreed 😉👍

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -65,12 +68,103 @@ class ServiceLocatorTests : XCTestCase {
...
+        do {
+
+            // Register the first service
+            try serviceLocator.register(service: testService)
+
+            // Register the same service, should fail with duplicateService
+            try serviceLocator.register(service: testService)
+
+        } catch ServiceLocatorError.duplicateService(_) {
+            print("✅"); return
```
p4checo: Here we could test the returned error itself and check if the associated value (the conflicting service name) matches `serviceName`.

On another point: do we really need to print something here? The test will pass if we simply leave this `catch` block empty 😉 

On a more 🚓 mode: dat `;` and think we don't need the `return` 😜 

Finally, we have a lot of these cases on the test suite, so the same comment applies to all

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -65,12 +68,103 @@ class ServiceLocatorTests : XCTestCase {
...
+        let testServiceLazy: () -> (String) = {
+            return "Test Register Service"
+        }
+
+        do {
+
+            // Register the first service (which should be lazy)
+            try serviceLocator.register(testServiceLazy)
+
+                // Register the a normal service, should fail with duplicateLazyService
```
p4checo: bad alignment 🚓

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -116,13 +210,110 @@ class ServiceLocatorTests : XCTestCase {
...
+
+        do {
+
+            // Register the first lazy service
+            try serviceLocator.register(testLazyServiceString)
+
+            // Force serviceLocator to call the lazy closure
+            let _: String = try serviceLocator.get()
+
+            // Register a early service, should fail with duplicateService
```
p4checo: What's an "early service" ? 🤔 😆

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -116,13 +210,110 @@ class ServiceLocatorTests : XCTestCase {
...
+            try serviceLocator.register(name: serviceName, testLazyServiceString)
+
+        } catch ServiceLocatorError.duplicateLazyServiceInit(_) {
+            print("✅")
+        } catch {
+            XCTFail("💥: Lazy Register failed with error: \(error)")
+        }
+    }
+
+    func testLazyRegisterWithNormal_UsingTypeInference_ShouldFailWithDuplicateService() {
```
p4checo: This name is a bit confusing to me, especially the `"WithNormal" part.

If I understand correctly, this case tests whether a service registered as lazy using type inference that then gets initialised (and thus "updated" to normal ) should fail with `duplicateService` (and not `duplicateLazyService`) when another register attempt is made with the same name.

If this is correct, why not call these tests something like:

`testLazyRegister_UsingTypeInferenceAndInitialisingIt_ShouldFailWithDuplicateService`

What do you say?

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -116,13 +210,110 @@ class ServiceLocatorTests : XCTestCase {
...
+            try serviceLocator.register(service: testService)
+
+        } catch ServiceLocatorError.duplicateService(_) {
+            print("✅")
+        } catch {
+            XCTFail("💥: Lazy Register failed with error: \(error)")
+        }
+    }
+
+    func testLazyRegisterWithNormal_UsingName_ShouldFailWithDuplicateService() {
```
p4checo: Same as before

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -135,8 +326,8 @@ class ServiceLocatorTests : XCTestCase {
             
             
             // Get registered services
-            let rStringService: String = try serviceLocator.get()
-            let rDoubleService: Double = try serviceLocator.get()
+            let rStringService: String = try serviceLocator.get(name: "String")
```
p4checo: I don't agree with these `get`'s receiving a name, because the services aren't registered using a name and the caller *shouldn't* know what name is auto generated by the SL.

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -189,11 +380,88 @@ class ServiceLocatorTests : XCTestCase {
...
         }
     }
+
+    // MARK: Failure
+
+    func testGet_UsingTypeInference_ShouldFailWithInexistingService() {
+
+        do {
+
+            // Check if there is a service with the same type
```
p4checo: This comment seems incorrect

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -189,11 +380,88 @@ class ServiceLocatorTests : XCTestCase {
...
+        }
+    }
+
+    func testGet_UsingName_ShouldFailWithInexistingService() {
+
+        let serviceName = "TestServiceName"
+
+        do {
+
+            // Check if there is a service with the same type
```
p4checo: Same as above

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -189,11 +380,88 @@ class ServiceLocatorTests : XCTestCase {
...
+        let serviceName = "Double"
+
+        do {
+
+            // Register a String type service
+            try serviceLocator.register(name: serviceName, service: testService)
+
+            // Get a Double type service
+            let _: Double = try serviceLocator.get()
+
```
p4checo: extra newline 🚓

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -189,11 +380,88 @@ class ServiceLocatorTests : XCTestCase {
...
+        let serviceName = "TestServiceName"
+
+        do {
+
+            // Register a String type service
+            try serviceLocator.register(name: serviceName, service: testService)
+
+            // Get a Double type service
+            let _: Double = try serviceLocator.get(name: serviceName)
+
```
p4checo: same as above 🚓

AlicerceTests/ServiceLocatorTests.swift
portellaa: 👦 🔎

AlicerceTests/ServiceLocatorTests.swift
portellaa: 👏 🥇 ✅

AlicerceTests/ServiceLocatorTests.swift
portellaa: 👮

AlicerceTests/ServiceLocatorTests.swift
portellaa: 😂 poor guy 😄

AlicerceTests/ServiceLocatorTests.swift
portellaa: Updated to use new api 🐭 💯

AlicerceTests/ServiceLocatorTests.swift
portellaa: PO 👜 😄

portellaa:
hey 🚓 addicted aka @p4checo, check my the last commit

portellaa:
👮‍♀️  now it's ready 😝

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -189,11 +383,129 @@ class ServiceLocatorTests : XCTestCase {
...
+        } catch ServiceLocatorError.serviceTypeMismatch(expected: let expectedType, found: let foundType) {
```
p4checo: This could be a single `catch let (...)` 💄

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -189,11 +383,129 @@ class ServiceLocatorTests : XCTestCase {
...
+            XCTAssertTrue(type(of: foundType) != type(of: expectedType))
```
p4checo: As discussed, using `type(of:)` on an `X.Type` will give `X.Type.Type`, aka "typeception" 😆 

I've made some tests, and this seems to work:
```swift

(...)
        } catch let ServiceLocatorError.serviceTypeMismatch(expected: expectedType, found: foundType) {
            XCTAssert(expectedType != foundType)
            XCTAssert(expectedType == Double.self)
            XCTAssert(foundType == type(of: testService))
        }
(...)
```
Apparently `XCTAssertTrue` doesn't work on type comparisons (as well as `XCTAssertNotEqual` and `XCTAssertEqual`) 🤷‍♂️. Type comparison must use some special equality operators, I guess. 

I also added some more checks on the types themselves 😉

AlicerceTests/ServiceLocatorTests.swift
portellaa: hey 👮‍♀️ this is fixed 😝

AlicerceTests/ServiceLocatorTests.swift
portellaa: hey 👮‍♀️ this is fixed 😝

AlicerceTests/ServiceLocatorTests.swift
portellaa: yeah i know, but i don't have a formalised idea which to use. but i will change, i also prefer the `let` before.

AlicerceTests/ServiceLocatorTests.swift
portellaa: hum, ok i was using `true` or `equal` hum strange.

portellaa: regarding the test with double, i'm not sure if i agree with this one. We could besides that validate that the type is not the same as the `testService`

portellaa:
New commit. Check the comments and see if you agree with this.

```
catch let ServiceLocatorError.serviceTypeMismatch(expected: expectedType, found: foundType) {
      let typeOfTestService = type(of: testService)
      XCTAssert(expectedType != foundType)
      XCTAssert(expectedType != typeOfTestService)
      XCTAssert(foundType == typeOfTestService)
}
```

AlicerceTests/ServiceLocatorTests.swift
p4checo: The equality test on `expectedType` (`Double` in this case) is to test if the error type itself is well built, so I think it adds value.

By only testing against `testService`'s type, we aren't testing this.

Besides, the test `expectedType != type(of: testService)` can already be logically implied from the other two tests:

`expectedType != foundType` **and**  `expectedType == type(of: testService)` **implies** `foundType != type(of: testService)` 🤓 

Don't you agree?

portellaa:
@p4checo i can't see your comment here 🤔 

but given that `testService` is a `String`, compare that the `expectedType` is not a `String` type but a different type. like a `Double` for example.

Both tests are valid, one is more explicit that the other. I was doing this when i got called to play  ⚽️ 🔨  😝 

```swift
private func service<T>(service: T, assertExpectedType expected: Any.Type, found: Any.Type) {
    XCTAssert(expected != found)
    XCTAssertTrue((expected is T) == false)
}
```

portellaa:
@p4checo squashed

p4checo:
Great work! 😄🍻

PR #5, opened 2017-03-08
Enhancements to ServiceLocator

p4checo:
Enhancements
- Synchronise access to `ServiceLocator` using a recursive lock
- Improve `unregister` API and internal helpers
- Unify `ServiceLocator` services into single dictionary
- Return service name on `ServiceLocator`’s `register`


portellaa:
On a first review, it seems very nice. I will take a better look later with fresh 👀 

I'll enable 🚓  😂

Alicerce/Sources/Utils/ServiceLocator.swift
```diff
@@ -8,104 +8,155 @@
...
         let name = buildName(for: Service.self, serviceName)
-        if let lazyServiceClosure = lazyServiceInits[name] {
-            let service: Service = try checkIfServiceValid(lazyServiceClosure())
-            
-            // Remove closure since the service is now validated
-            lazyServiceInits[name] = nil
-            
-            // This should *never* fail unless we have some name collision bug, so 💥 immediately
-            try! register(name: name, service: service)
+        var service: Service?
```
portellaa: We could use ! directly if we know that the variable will be populated. Or do you want to fail if it don't?

This should be a validate responsibility, i think.

Alicerce/Sources/Utils/ServiceLocator.swift
```diff
@@ -8,104 +8,155 @@
...
+
+        try synchronized {
+            service = try locate(name)
         }
-        
-        guard let anyService = services[name] else {
-            throw ServiceLocatorError.inexistentService
+
+        // This should *never* fail unless we have some bug returning the service, so 💥 immediately
+        return service!
```
portellaa: If you like the suggestion, you can remove the force here. We should trust that validate does all the validation required 💪 😄

Alicerce/Sources/Utils/ServiceLocator.swift
p4checo: Well spotted! That didn't occur to me at the time 😛 Will Change

p4checo:
Ready for re-review

p4checo:
Thanks 🙇

PR #7, opened 2017-03-28
Register Cell Helpers

jakewelton:
# Register Cell Helpers
This adds 2 functions. One to the `UITableView` extension and another to the `UICollectionView` extension. It basically just makes registering a cell simpler when the cell conforms to `ViewCellProtocol`.

### DIFF
--- a/Alicerce/Sources/Extensions/UICollectionView.swift
+++ b/Alicerce/Sources/Extensions/UICollectionView.swift
@@ -18,4 +18,12 @@ public extension UICollectionView {
         
         return cell
     }
+    
+    public func register<T: UICollectionViewCell>(_ cellType: T.Type) where T: ViewCellProtocol {
+        register(cellType, forCellWithReuseIdentifier: cellType.reuseIdentifier)
+    }
+    
+    public func register<T: AnyObject>(_ viewType: T.Type, forSupplementaryViewOfKind kind: String) where T: ViewCellProtocol {
+        register(viewType, forSupplementaryViewOfKind: kind, withReuseIdentifier: viewType.reuseIdentifier)
+    }
 }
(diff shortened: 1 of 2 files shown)


portellaa:
Great work @jakewelton 👍

p4checo:
My only suggestion would be to squash these 3 commits before merging 😉

jakewelton:
@portellaa Good idea. Commit 9862c078c954153270ff47c29955f0e725bc1dad now includes the supplementary view registration too (although I'm not 100% happy with the naming, so any ideas are welcome 😄). 

@p4checo Should I squash the commits before the merge? Or just use the squash and merge function on GitHub?

Alicerce/Sources/Extensions/UICollectionView.swift
```diff
@@ -22,4 +22,8 @@ public extension UICollectionView {
     public func register<T: UICollectionViewCell>(_ cellType: T.Type) where T: ViewCellProtocol {
         register(cellType, forCellWithReuseIdentifier: cellType.reuseIdentifier)
     }
+    
+    public func register<T: AnyObject>(_ viewType: T.Type, forSupplementaryViewOfKind kind: String) where T: ViewCellProtocol {
+        register(viewType.self, forSupplementaryViewOfKind: kind, withReuseIdentifier: viewType.reuseIdentifier)
```
p4checo: I don't think we need `.self` here (and below either)

p4checo:
@jakewelton however you prefer. I've personally never used GitHub's integrated squashing feature, so I don't know how it works. The most important thing is to be able to edit the commit message for the final commit 😛

Alicerce/Sources/Extensions/UICollectionView.swift
jakewelton: You're right. Not sure why I added that in 😕 
I've removed in 3b1d945bb496e940352f908ef2ee9655dfaed540.

jakewelton:
@p4checo @portellaa All done. I've squashed the commits too now 😄

portellaa:
Well done 🎉 👍 🥇

PR #8, opened 2017-04-02
Improve README

portellaa:
Created getting started section for how to include the framework into the project

TODO list with some items to be done recently

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,37 @@
 # Alicerce
 A base for iOS Applications with ❤️ from Mindera 🤠
+
+[![Swift 3.0](https://img.shields.io/badge/Swift-3.0-orange.svg?style=flat)](https://developer.apple.com/swift/)
+[![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
+[![Build Status](https://travis-ci.org/Mindera/Alicerce.svg?branch=master)](https://travis-ci.org/Mindera/Alicerce)
+[![Codecov branch](https://img.shields.io/codecov/c/github/mindera/Alicerce/master.svg)]()
+[![Github All Releases](https://img.shields.io/github/downloads/mindera/Alicerce/total.svg)]()
+
+
+## Getting started
+
+Alicerce supports iOS 9.0+ 📱 (for now).
+
+#### Carthage
+
+If you use [Carthage][https://github.com/Carthage/Carthage] to manage your dependencies, simply add
+Alicerce to your `Cartfile`:
+
+```
+github "Mindera/Alicerce"
+```
+
+#### Swift Package Manager
+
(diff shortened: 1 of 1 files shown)


p4checo:
Please rebase 🔧

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,695 @@
...
+            let rStringService: String = try serviceLocator.get()
+            let rDoubleService: Double = try serviceLocator.get()
+            let rIntService: Int = try serviceLocator.get()
+            
+            // Validate registered services
+            XCTAssertEqual(rStringService, testServiceString)
+            XCTAssertEqual(rDoubleService, testServiceDouble)
+            XCTAssertEqual(rIntService, testServiceInt)
+            
+        } catch let error {
```
jakewelton: I don't think you need to declare the error variable here, as its now added by default.

AlicerceTests/ServiceLocatorTests.swift
```diff
@@ -0,0 +1,695 @@
...
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import XCTest
+
+@testable import Alicerce
+
+class ServiceLocatorTests : XCTestCase {
+    
+    var serviceLocator = ServiceLocator()
```
jakewelton: This is only a minor thing, but with `setUp` running before each test, do you think we need to create an object here? Normally I opt for an implicitly unwrapped optional `var serviceLocator: ServiceLocator!`.

What do you think?

AlicerceTests/ServiceLocatorTests.swift
portellaa: Thanks @jakewelton, didn't know that.

AlicerceTests/ServiceLocatorTests.swift
portellaa: Sorry, i don't like implicit unwrapped so i avoid to use it all the time. but yeah, it makes sense.

portellaa:
@jakewelton this files moved from `AlicerceTests/Utils/ServiceLocatorTests.swift`

Sorry we are moving too fast with this, but we are working 200% on this with the new project.

Great work mate 🍺

AlicerceTests/ServiceLocatorTests.swift
jakewelton: Thats fair enough. I too normally hate implicitly unwrapped options, but sometimes make an exception for unit tests as they are not production code. But thats fair enough 👍

portellaa:
Hi @jakewelton

I did change the tests to do what you suggested, it makes a lot Of sense.

Sorry to not wait at this stage, but we really need to move forward with this part.

:beers_mug:

jakewelton:
@portellaa That's fine, don't worry 😄 👍 However, I can't see the unit tests in the diff anymore for this PR. Were they moved to another PR?

PR #9, opened 2017-04-03
Add CoreDataStack 💾

p4checo:
- Created `CoreDataEntity` protocol to contain some utilities for common
operations which aren’t yet available natively on iOS 8 and 9.
- Created `CoreDataStack` protocol to represent a Core Data Stack,
together with a set of extensions to create the required
components of a Core Data stack, as well as methods for basic CRUD
operations on a generic `CoreDataEntity`.
- Created `NestedContextCoreDataStack` for a nested context Core Data
stack implementation, a `MainQueueNestedCoreDataStack` subclass and a
`PrivateQueueNestedCoreDataStack` type alias.
- Created `DualContextCoreDataStack` for a nested context Core Data
stack implementation and a `MainQueueDualCoreDataStack` type alias.
- Created `ManagedObjectReflectable` protocol to enable the reflection
of one type to another.
- Created a `CoreDataStack+ManagedObjectReflectable` protocol extension
to specialise CRUD operations for any `ManagedObjectReflectable` and
abstract the reflection logic between both Entities automatically ✨.
- Created Unit Tests for `CoreDataStack` setup and operation methods.
- Created `MockErrorManagedObjectContext` to mock errors on a context.
- Created extension methods on `NSManagedObjectContext`
`performThrowing` and `performThrowingAndWait` which are throwing
versions of `perform` and `performAndWait` that also return a generic
value `T` 💪


Alicerce/Sources/Persistence/NSManagedObjectContext+CoreDataStack.swift
```diff
@@ -0,0 +1,95 @@
...
+        perform {
+            do {
+                let value = try block()
+                completion(value, nil)
+            }
+            catch { completion(nil, error) }
+        }
+    }
+
+    func performThrowingAndWait<T>(_ block: @escaping () throws -> T) throws -> T {
```
portellaa: What do you think of a closure for the `block`?

Alicerce/Sources/Persistence/SiblingContextCoreDataStack.swift
```diff
@@ -0,0 +1,128 @@
...
+
+            NotificationCenter.default.addObserver(self,
+                                                   selector: #selector(backgroundContextChanged),
+                                                   name: .NSManagedObjectContextDidSave,
+                                                   object: backgroundContext)
+        }
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
```
portellaa: 🚔 You could remove this, since we are only supporting iOS9. But it's up to you 😄 😝

Alicerce/Sources/Persistence/SiblingContextCoreDataStack.swift
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private static let defaultStoreLoadCompletionHandler: (Any, Error?) -> Void = { (store, error) in
+        if let error = error {
+            fatalError("💥: Failed to load persistent store \(store)! Error: \(error)")
+        }
+    }
+
+    func context(withType type: CoreDataStackContextType) -> NSManagedObjectContext {
```
portellaa: i would replace `withType` with `with` 😝 (this is just to make noise) 🛎

p4checo: I had it like that before, but then found it weird when invoking and changed it to its current form. Examples:

`let context = coreDataStack.context(type: .work)`

vs 

`let context = coreDataStack.context(withType: .work)`

I find it more "readable" using the `withType` parameter 😛

Alicerce/Sources/Persistence/NSManagedObjectContext+CoreDataStack.swift
p4checo: I'll add it 😉

PR #12, opened 2017-04-04
Rename travis and fix language

portellaa:


p4checo:
For anyone interested, Travis was taking to long so we had to let it go... 😢

portellaa:
More detail on this, they already experienced some problems on this: https://github.com/travis-ci/travis-ci/issues/7304

PR #14, opened 2017-04-04
Service locator tests

portellaa:
* Removed error declaration from catch


portellaa:
@jakewelton moved the unit tests to here

PR #20, opened 2017-04-07
Optional then extension

portellaa:


AlicerceTests/Extensions/OptionalTests.swift
```diff
@@ -0,0 +1,22 @@
...
+
+final class OptionalTests: XCTestCase {
+
+    func testThen_UsingNonOptionalValue_ShouldUnwrapTheValue() {
+        let anOptionalString: String? = "😎"
+
+        anOptionalString.then { value in
+            XCTAssert(true, "🚀 \(value) unwrapped ✅")
+        }
+    }
```
jakewelton: Do you think there would be any merit in adding a test for when the string is nil?

portellaa: done @jakewelton

jakewelton: Nice 👍

AlicerceTests/Extensions/OptionalTests.swift
```diff
@@ -0,0 +1,32 @@
...
+
+@testable import Alicerce
+
+final class OptionalTests: XCTestCase {
+
+    func testThen_UsingNonOptionalValue_ShouldUnwrapTheValue() {
+        let anOptionalString: String? = "😎"
+
+        anOptionalString.then { value in
+            XCTAssert(true, "🚀 \(value) unwrapped ✅")
```
p4checo: This `XCTAssert` doesn't do anything and can be converted to a simple comment 😇

To test that the `then`closure effectively ran we can simply modify `anOptionalString` inside it and assert it's value after `then` is invoked 😉

AlicerceTests/Extensions/OptionalTests.swift
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    func testThen_UsingOptionalValue_ShoudNotUnwrapTheValue() {
+        let nullValue: String? = nil
+
+        nullValue.then { _ in
+            XCTAssert(false, "💥 nil unwrapped 😱")
```
p4checo: This would be better expressed as `XCTFail` 😄

AlicerceTests/Extensions/OptionalTests.swift
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        XCTAssert(true, "🚀 value not unwrapped ✅")
```
p4checo: Same as above (assert always passes, and we can add a check using the `nullValue`'s own value)

portellaa:
@p4checo can you check the changes?

AlicerceTests/Extensions/OptionalTests.swift
```diff
@@ -0,0 +1,30 @@
...
+
+@testable import Alicerce
+
+final class OptionalTests: XCTestCase {
+
+    func testThen_UsingNonOptionalValue_ShouldUnwrapTheValue() {
+        let anOptionalString: String? = "😎"
+
+        anOptionalString.then { value in
+            XCTAssertTrue(anOptionalString! == value, "🚀 \(value) unwrapped ✅")
```
p4checo: Sorry for the delay, @portellaa 😊 

What I suggested was slightly different from this. Even though this is a little bit better, we still don't know for sure that the block was effectively executed, right?

Some notes to fix that:
1. We can keep an assertion inside the `then` closure, like: 
`XCTAsserEqual(anOptionalString, value, "🔥: \(value) not unwrapped! 😱")`
(note that the string in this case is the message shown when the assert *fails* 😛 )

2. Change `value` *inside* the `then` closure, e.g.:  
`anOptionalString = "😇"`
3. Do an assert *after* the `then` closure to check that the value was properly modified, e.g.: 
`XCTAsserEqual(anOptionalString, "😇", "🔥: `then` closure not executed! 😭")`

portellaa: 🤔 u'r right, we don't. 

1. ok, i'm not used to `XCTest`, for me AssertTrue and AssertEqual are the same, but i understand the differences

2. 👍 

3. 👍

p4checo: Yeah, they can be a bit confusing at first, but then you get the hang of it 😉 

As for `XCTAssertTrue`, on these cases I simply use `XCTAssert` which is the same 😛

portellaa: @p4checo done 👍

PR #22, opened 2017-04-11
Simple Network 📚 with URLSession implementation

portellaa:
* Protocol Mappable for model parsing
* NetworkStack protocol
* URLSessionNetworkStack implementation
* Network module with network related objects

**UnitTests**
* URLSessionNetworkStack
* Parse
* Resource
* Mappable

⚠️ I changed the Unit tests folder structure to contain source code files and resource files ⚠️


Alicerce/Sources/Network/Mappable.swift
```diff
@@ -0,0 +1,41 @@
...
+    /// - Throws: 
+    /// A MappableError
+    static func model(from object: Any) throws -> Self
+
+    /// Converts Itself into a `AnyObject` (Dictionary) representation
+    /// You can simply return `NSNull()` if it doesn't make sense in your context to do that.
+    ///
+    /// - Returns: 
+    /// An `AnyObject` (like a `Dictionary` representation)
+    func json() -> AnyObject
```
jakewelton: I'm in two minds about what I'm about to suggest. I'll suggest it anyway :smile: What do you think about making this a property? Could be a computed property when implemented? Just thinking it might look cleaner at the call site? 🤔

Alicerce/Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -0,0 +1,60 @@
...
+                                        delegate: nil,
+                                        delegateQueue: configuration.delegateQueue)
+
+            self.init(baseURL: configuration.baseURL, session: urlSession)
+        }
+
+        public func fetch<R: NetworkResource>(resource: R, _ completion: @escaping Network.CompletionClosure) {
+            let request = resource.toRequest(withBaseURL: baseURL)
+
+            session.dataTask(with: request) { (data, response, error) in
```
jakewelton: I believe you don't need the brackets around the closure arguments :smile:

Alicerce/Sources/Resource/NetworkResource.swift
```diff
@@ -0,0 +1,47 @@
...
+
+        urlRequest.allHTTPHeaderFields = headers
+        urlRequest.httpBody = body
+        urlRequest.httpMethod = method.rawValue
+
+        return urlRequest
+    }
+
+    private func buildQueryItems() -> [URLQueryItem]? {
+        guard let query = query, query.isEmpty == false else {
```
jakewelton: This is just personal preference, but normally I use `!query,isEmpty` rather than `query.isEmpty == false` :)

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,192 @@
...
+    var mockDataTaskData: Data? = Data()
+    var mockDataTaskError: Error? = nil
+    var mockURLResponse: URLResponse = URLResponse()
+
+    override func dataTask(with request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask {
+
+        let dataTask = MockURLSessionDataTask()
+
+        dataTask.resumeInvokedClosure = { [weak self] in
+            guard let `self` = self else { fatalError("🔥: `self` must be defined!") }
```
jakewelton: I could be wrong, but isn't the use of `'self' = self` going to be removed soon, as its a hack? Should we preempt this and use `strongSelf` instead?

Alicerce/Sources/Network/Mappable.swift
portellaa: Sorry, but i don't agree with make all computed properties if they don't have input parameters.

In the computer science terms, this makes sense to be a function because it can have logic assigned to it, it depends on the class that implements the method to decide what to do.

But this is just my point of view, i'm ok to change if it doesn't makes sense to you and the other guys.

But, let's not make all the `func` that doesn't have parameters as computed properties.

jakewelton: Thats why I was in two minds about it. Part of me thought it would go well as a computed property due to 0 arguments, however it does act like a function, which means maybe a property is not the way. I'm happy with keeping it as a function :smile:

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
portellaa: it is 😢

Alicerce/Sources/Resource/NetworkResource.swift
portellaa: i don't have a personal choice, i like both, but this one is more declarative

Alicerce/Sources/Network/URLSessionNetworkStack.swift
portellaa: i don't, it was the autocomplete and the bed calling me 😂

portellaa:
@jakewelton updated with your review 😉

Alicerce/Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+//  Created by Luís Afonso on 07/04/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public extension Network {
+
+    public final class URLSessionNetworkStack: NetworkStack {
```
p4checo: no need for `public` 💄

Alicerce/Sources/Network/Network.swift
```diff
@@ -0,0 +1,42 @@
...
+    // MARK: - TypeAlias
+
+    public typealias CompletionClosure = (_ inner: () throws -> Data) -> Void
+
+    // MARK: - Network Error
+
+    enum Error: Swift.Error {
+        case http(code: HTTP.StatusCode, description: String?)
+        case noData
+        case url(Swift.Error)
```
p4checo: Not sure about naming this error case `url`, but can't think of a nicer alternative... 🤔

The first thing that comes to mind is `network` 😆, and these errors will be stuff like disconnected, timeout, etc

Alicerce/Sources/Network/Parse.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// A Parse.Error that can be of type: 
+    ///   - `serialization` if json serialization failed
+    ///   - `mapping` if model mapping failed
+    ///   - `other` an unknown error
+    static func json<T: Mappable>(data: Data) throws -> T {
+
+        let decodedObject: T
+
+        do {
+            // Check if we can serialize the data into a Dictionary
```
p4checo: Might not be a Dictionary

Alicerce/Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -0,0 +1,66 @@
...
+                    return completion { throw Network.Error.url(error) }
+                }
+
+                guard let urlResponse = response as? HTTPURLResponse else {
+                    return completion { throw Network.Error.badResponse }
+                }
+
+                let httpStatusCode = HTTP.StatusCode(urlResponse.statusCode)
+
+                guard case HTTP.StatusCode.success(_) = httpStatusCode else {
```
p4checo: can be simplified to `guard case .success = httpStatusCode else {`

Alicerce/Sources/Network/Mappable.swift
p4checo: On another topic, I think `NSJSONSerialization` now uses `Any` on its API's, so we can simply return `Any` instead of `AnyObject` :wink:

AlicerceTests/Sources/Network/MappableTestCase.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    func testModel_WhenInputObjectIsValid_ItShouldReturnAFilledModel() {
+        let aValidDict = [
+            "data" : "👍"
+        ]
+
+        do {
+            let mappedModel = try MappableModel.model(from: aValidDict)
+
+            XCTAssertEqual(mappedModel, MappableModel(data: "👍"), "✅ it created a valid model")
```
p4checo: You keep using the assertion's **failure** messages to give **success** messages... 😂

AlicerceTests/Sources/Network/MappableTestCase.swift
```diff
@@ -0,0 +1,62 @@
...
+        } catch {
+            XCTFail("🔥 It thrown an invalid error 👉 \(error) 😱")
```
p4checo: Grammar 👮 : "An invalid error was thrown"

Also, I would call it "unexpected" instead of "invalid" 😝

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,192 @@
...
+        let mockData = "🎉".data(using: .utf8)
+        mockSession.mockDataTaskData = mockData
+
+        let network = Network.URLSessionNetworkStack(baseURL: baseURL, session: mockSession)
+
+        network.fetch(resource: resource) { (inner: () throws -> Data) -> Void in
+            do {
+                let data = try inner()
+
+                XCTAssertEqual(data, mockData, "✅ received same data from network 😎")
```
p4checo: same as before 😛

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,192 @@
...
+
+        let network = Network.URLSessionNetworkStack(baseURL: baseURL, session: mockSession)
+
+        network.fetch(resource: resource) { (inner: () throws -> Data) -> Void in
+            do {
+                let _ = try inner()
+
+                XCTFail("🔥 should throw an error 🤔")
+            } catch let Network.Error.url(receivedError as NSError) {
+                XCTAssertEqual(receivedError, mockError, "✅ received same error from network 🤠")
```
p4checo: Same as above 😛

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,192 @@
...
+            do {
+                let _ = try inner()
+            } catch _ {}
+        }
+    }
+
+    func testURLSession_WhenResponseIsSuccessful_ItShouldCallCompletionClosureWithData() {
+        let baseURL = URL(string: "http://")!
+
+        let mockSession = MockURLSession()
```
p4checo: We could move this `mockSession` to a var on the test itself and reset it on each `setUp`/`tearDown`. It would avoid all this duplicated setup on multiple tests 😉

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,192 @@
...
+
+// MARK: - Network Mocks
+
+final class MockURLSession : URLSession {
+
+    var mockDataTaskData: Data? = Data()
+    var mockDataTaskError: Error? = nil
+    var mockURLResponse: URLResponse = URLResponse()
+
+    override func dataTask(with request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask {
```
p4checo: Please break this into multiple lines 🚓

Alicerce/Sources/Network/Mappable.swift
portellaa: @p4checo yes, this is my port from 2.3 and i forgot to update it. sorry and thanks

Alicerce/Sources/Network/Mappable.swift
```diff
@@ -0,0 +1,41 @@
...
+    var description: String {
+        switch self {
+        case let .custom(description): return description
+        }
+    }
+}
+
+public protocol Mappable {
+
+    /// Converts the object from a `Any` (Dictionary) to Itself.
```
portellaa: the doc is wrong. it can be anything, i use to apply the `Mappable` to everything, not only for dictionaries and it can be an array, yes.
i will fix the doc.

Alicerce/Sources/Network/Network.swift
portellaa: what about `response`? 😕

Alicerce/Sources/Network/Parse.swift
```diff
@@ -0,0 +1,62 @@
...
+//  Alicerce
+//
+//  Created by Luís Afonso on 06/04/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import UIKit
+
+public enum Parse {
+    indirect enum Error: Swift.Error {
```
portellaa: of course not. don't know how it appeared 😕

Alicerce/Sources/Network/Parse.swift
portellaa: this log is just stupid. is more than obvious what it does 🙄

Alicerce/Sources/Network/Network.swift
```diff
@@ -0,0 +1,42 @@
...
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public enum Network {
+
+    // MARK: - TypeAlias
+
+    public typealias CompletionClosure = (_ inner: () throws -> Data) -> Void
```
portellaa: i think we do, otherwise it will be `internal` and cannot be used by classes that calls `fetch`

Alicerce/Sources/Network/URLSessionNetworkStack.swift
portellaa: 💄

Alicerce/Sources/Network/Network.swift
p4checo: You are probably right, since inside it will default to `internal` again 😅

Alicerce/Sources/Network/Network.swift
p4checo: Don't like `response` very much either 😞

Alicerce/Sources/Network/Parse.swift
```diff
@@ -0,0 +1,61 @@
...
+    ///   - `mapping` if model mapping failed
+    ///   - `other` an unknown error
+    static func json<T: Mappable>(data: Data) throws -> T {
+
+        let decodedObject: T
+
+        do {
+            let json = try JSONSerialization.jsonObject(with: data, options: [])
+
+            decodedObject = try T.model(from: json)
```
p4checo: I think we can return directly from inside the `do..catch` block, so no need for the external var at the beginning of the func 😉

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+@testable import Alicerce
+
+final class URLSessionNetworkStackTestCase: XCTestCase {
+
+    private var mockSession: MockURLSession!
+
+    override func setUp() {
+        mockSession = MockURLSession()
+    }
+
```
p4checo: Why no `tearDown` to free the `mockSession`? 😢

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+import XCTest
+
+@testable import Alicerce
+
+final class URLSessionNetworkStackTestCase: XCTestCase {
+
+    private var mockSession: MockURLSession!
+
+    override func setUp() {
+        mockSession = MockURLSession()
```
p4checo: You forgot to call `super.setUp()`

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+    func testURLSession_WhenResponseIsNotHTTPKind_ItShouldThrowABadResponseError() {
+        let baseURL = URL(string: "http://")!
+
+        let network = Network.URLSessionNetworkStack(baseURL: baseURL, session: mockSession)
+
+        network.fetch(resource: resource) { (inner: () throws -> Data) -> Void in
+            do {
+                let _ = try inner()
+
+                XCTFail("🔥 should throw an error 🤔")
```
p4checo: "should not throw"?

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+                                                      headerFields: nil)!
+        mockSession.mockDataTaskData = nil
+
+        let network = Network.URLSessionNetworkStack(baseURL: baseURL, session: mockSession)
+
+        network.fetch(resource: resource) { (inner: () throws -> Data) -> Void in
+            do {
+                let _ = try inner()
+
+                XCTFail("🔥 should throw an error 🤔")
```
p4checo: Same

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+                                                      httpVersion: nil,
+                                                      headerFields: nil)!
+
+        let network = Network.URLSessionNetworkStack(baseURL: baseURL, session: mockSession)
+
+        network.fetch(resource: resource) { (inner: () throws -> Data) -> Void in
+            do {
+                let _ = try inner()
+
+                XCTFail("🔥 should throw an error 🤔")
```
p4checo: Same as above

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
portellaa: no no, it should throw

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
portellaa: this is just to fail in case if an error is not thrown

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
portellaa: same 😂

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
p4checo: You're right 😅

AlicerceTests/Sources/Network/URLSessionNetworkStackTestCase.swift
portellaa: fixed

portellaa:
@jakewelton can we merge this?

PR #25, opened 2017-04-13
Fix ignore on codecov.yml and specify packages on travis.yml

portellaa:


.codecov.yml
```diff
@@ -29,7 +29,8 @@ coverage:
                                  # options: success, error, failure
         if_not_found: success
         if_ci_failed: error
-
+  
   ignore:          # files and folders that will be removed during processing
-    - "AlicerceTests/*"
+    - AlicerceTests*
+    - *.xctest
```
p4checo: Do we still need this? It didn't have any effect on our tests since codecov looks at the entire `Coverage.profdata` file which in our case has coverage for both targets (Alicerce.framework and AlicerceTests.xctest)

.travis.yml
```diff
@@ -16,6 +16,6 @@ script:
   - set -o pipefail
   - xcodebuild $XCODE_ACTION -project "$TRAVIS_XCODE_PROJECT" -scheme "$XCODE_SCHEME" -sdk "$XCODE_SDK" -destination "$XCODE_DESTINATION" -derivedDataPath "$XCODE_DERIVED_DATA_PATH"
 after_success:
-  - bash <(curl -s https://codecov.io/bash) -D $XCODE_DERIVED_DATA_PATH
+  - bash <(curl -s https://codecov.io/bash) -D $XCODE_DERIVED_DATA_PATH -J "^Alicerce$"
```
p4checo: What do you think of making this more dynamic by using an env var, like `XCODE_TARGET_NAME`?

portellaa: good suggestion 👮‍♀️

PR #26, opened 2017-04-17
Require from John Sundell

portellaa:
I was about to do something similar to this, but i read an article these days from John Snudell where he talked about his method and i decide to port it to here.

If you don't agree with this, please reject or shout any suggestions 😉

### DIFF
--- a/Alicerce.xcodeproj/project.pbxproj
+++ b/Alicerce.xcodeproj/project.pbxproj
@@ -470,7 +470,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0820;
-				LastUpgradeCheck = 0810;
+				LastUpgradeCheck = 0830;
 				ORGANIZATIONNAME = Mindera;
 				TargetAttributes = {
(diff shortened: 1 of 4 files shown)


p4checo:
We should find a way to test these crashing error cases so that the coverage isn't so low 🤔

portellaa:
Yes. I was trying to do a magic with preconditionFailure, like change the method implementation to our one in the tests target. I saw a gist doing that for precondition only, but it didn't work. It "`override`" the method and use expectations to validate if it was called.
I will try that again, but i think that didn't work because this is a `Never` function.

But, i think we don't need to be so "aggressive" in the tests, we are testing the functionality of unwrap the optional and we have to make sure that this doesn't breaks.
At least, part of me see the tests like that 😝 (i think the books says that)

I can take a look again into that.

p4checo:
Yeah, it sucks that we don't have a simple way of catching these traps on our unit tests. 

While some cases aren't supposed to happen and might even be almost impossible to reach, some others are reachable, but essentially untestable 😓

This still looks good to me, I just wish we could have a higher coverage by being able to test more cases 😉

PR #28, opened 2017-04-17
Add initial Deep Linking infrastructure

p4checo:
- Created `ApplicationRouter` protocol to handle `UIApplicationDelegate`
external route entry points.
- Created `ApplicationRoute` enum to represent a type of route received
by the `AppDelegate`.
- Created `Route` careless enum to namespace route specific types:
  + Created `Route.Component` enum to represent the different types of
path components in a route.
  + Created `Route.Tree` enum to represent a routing tree to efficiently
perform matching against.
- Created `Router` protocol to handle registration/unregistration/handle
of URL routes.
- Created `RouteHandler` protocol to represent a handler for a specific
route.
- Created `TreeRouter` that handles URL routes using an internal
`Route.Tree`.
- Added Unit Tests for `Route.Tree`

### DIFF
--- a/Alicerce/Sources/DeepLinking/Route+Tree.swift
+++ b/Alicerce/Sources/DeepLinking/Route+Tree.swift
@@ -8,13 +8,13 @@
 
 import Foundation
 
-extension Route {
+public extension Route {
 
-    indirect enum Tree<Handler> {
+    public indirect enum Tree<Handler> {
 
         // MARK: Nested types
 
-        enum Error: Swift.Error {
+        public enum Error: Swift.Error {
             case invalidRoute
             case duplicateEmptyComponent
             case conflictingParameterName(existing: String, new: String)
@@ -22,13 +22,13 @@ extension Route {
             case invalidComponent(Component)
         }
 
-        enum Edge {
+        public enum Edge {
             case simple(Tree<Handler>)
             case parameter(String?, Tree<Handler>)
         }
 
-        typealias ChildEdges = [Component.Key : Edge]
-        typealias Match = (parameters: [String : String], handler: Handler)
+        public typealias ChildEdges = [Component.Key : Edge]
+        public typealias Match = (parameters: [String : String], handler: Handler)
 
         // MARK: Implementation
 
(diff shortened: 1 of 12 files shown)


Alicerce/Sources/DeepLinking/Router.swift
```diff
@@ -0,0 +1,126 @@
...
+
+final class TreeRouter<Handler: RouteHandler>: Router {
+
+    typealias Match = Route.Tree<Handler>.Match
+    private typealias ParsedRoute = (scheme: Route.Scheme, components: [Route.Component])
+
+    enum Error: Swift.Error {
+        case invalidRoute(Swift.Error)
+        case routeNotFound
+        //case routeAlreadyHandled(route: [Component], handler: Handler)
```
mschuetz72: would it make sense to remove this commented line, or should we prefix it with something like //TODO: ?

p4checo: I forgot to remove this 😅 

It has been fixed, since more errors have been added 😉

Alicerce/Sources/DeepLinking/Route+Tree.swift
```diff
@@ -0,0 +1,270 @@
...
+
+        private func matchParameterEdge(parameterName: String?,
+                                        childTree: Tree<Handler>,
+                                        currentComponent: Component,
+                                        remainingRoute: [Component]) throws -> Match {
+            let (parameters, handler) = try childTree.match(route: remainingRoute)
+             
+            guard let parameterName = parameterName else { return (parameters, handler) }
+
+            var allParameters = parameters
```
portellaa: 💄 we could do the same as in the `guard`, and redefine the variable: `var parameters = parameters` just an idea 😛

Alicerce/Sources/DeepLinking/Router.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    func route(_ route: URL) throws {
+        let (scheme, routeComponents, queryItems) = try parseRoute(route)
+
+        let match: Match = try queue.sync { [unowned self] in
+            guard let schemeTree = self.routes[scheme] else { throw Error.routeNotFound }
+
+            let match: Match
```
p4checo: I completely agree. Don't know where I got the idea that I couldn't `return` directly from inside `do...catch` scopes 😅

p4checo:
Thanks! 🙏 

Let's hope it works! 🤞 At least I tried to cover all the cases I could remember on the UT's 😆

PR #30, opened 2017-04-18
Support for UIColor hex init

portellaa:
Adds support to create an UIColor from hex RGB or RGBA.


Alicerce/Sources/View/UIColor.swift
```diff
@@ -0,0 +1,58 @@
...
+    private typealias Components = (red: CGFloat, green: CGFloat, blue: CGFloat, alpha: CGFloat)
+
+    public convenience init(hex: String) {
+        let hex = hex.trimmingCharacters(in: .whitespacesAndNewlines)
+            .replacingOccurrences(of: "#", with: "")
+
+        var hexValue: UInt32 = 0
+
+        guard Scanner(string: hex).scanHexInt32(&hexValue) else {
+            fatalError("😱 Cannot convert string into `UInt32`")
```
p4checo: Not sure about crashing here... 🤔 

Why not throw and error? 😛

Alicerce/Sources/View/UIColor.swift
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        let components: Components = {
+            switch hex.characters.count {
+            case 6:
+                return UIColor.components(fromHex6: hexValue)
+            case 8:
+                return UIColor.components(fromHex8: hexValue)
+            default:
+                fatalError("😱 hex size not supported 😇")
```
p4checo: Same as above 😉

Alicerce/Sources/View/UIColor.swift
portellaa: 😢 i know, but i don't like to throw here, i have to make those validations later in the code, which makes it "dirty".
i know, it seems more correct to throw, but i think we will not be changing these values in runtime and initialising `UIColor` with those values, at least not most of the times.

I like to have my colours like this:

```swift
extension UIColor {
    static var tabbarTintColor: UIColor { return UIColor(hex: "#e73a35") }
}
```

If i `throw`, i have to make it `return try! UIColor(hex: "#e73a35")` (because i know that this will not fail) 😝 

Honestly i agree with you, but i don't like to throw in this particular case 😢 but...

and looking into the Apple API, they don't throwable init for `UIColor`

p4checo: I also don't particularly like of throwing on these types of initialisers, because it requires more boilerplate when using, but I asked this question since API wise it would perhaps be the **right thing ®** 😄 

However, Apple also traps on invalid input (e.g. duplicate keys on dictionaries), so we might be safe to simply crash here,. which makes using the API much easier 🤔

PR #31, opened 2017-04-18
Support for persistence in memory and disk

portellaa:


Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,330 @@
...
+    private let cache = NSCache<NSString, NSData>()
+
+    private let configuration: Configuration
+
+    weak var delegate: DiskMemoryPersistenceDelegate?
+
+    private var diskCacheEnabled: Bool = false
+
+    private let readOperationQueue: OperationQueue = {
+        $0.name = "com.nap.mobile_mrp_mrporter.cache.disk.read"
```
mschuetz72: I'd reconsider this key 👀

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,330 @@
...
+        return $0
+    }(OperationQueue())
+    private let writeOperationQueue: OperationQueue = {
+        $0.name = "com.nap.mobile_mrp_mrporter.cache.disk.write"
```
mschuetz72: I'd reconsider this key 👀

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,330 @@
...
+            return false
+        }
+
+        return true
+    }
+
+    private func directoryContents(with keys: [URLResourceKey]) -> [URL] {
+        guard let url = URL(string: configuration.path) else {
+            assertionFailure("💥 failed to create url with path 👉 `\(configuration.path)`")
+
```
mschuetz72: Just 💄, but I think we can save this line

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,330 @@
...
+            return []
+        }
+
+        guard let urls = try? fileManager.contentsOfDirectory(at: url,
+                                                              includingPropertiesForKeys: keys,
+                                                              options: .skipsPackageDescendants)
+            else {
+                assertionFailure("💥 failed to read directory content for path 👉 \(url.absoluteString)")
+
```
mschuetz72: Just 💄, but I think we can save this line

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,330 @@
...
+                guard evictSize < extraOccupiedDiskSize else {
+                    return false
+                }
+
+                evictSize += $0.fileAttr.size
+
+                return true
+            }
+
+
```
mschuetz72: Just 💄, but I think we can save this line

AlicerceTests/Sources/Persistence/DiskMemoryPersistenceTestCase.swift
```diff
@@ -0,0 +1,235 @@
...
+//
+//            checkAndRetry(times: 5) { retry in
+//                if fileExists("Test2/mr-minder") {
+//                    retry()
+//                } else {
+//                    deleteItem("Test2")
+//                }
+//            }
+//        }
+//    }
```
mschuetz72: would it make sense to remove this commented code before merging to master?

AlicerceTests/Sources/Utils.swift
```diff
@@ -0,0 +1,42 @@
...
+import UIKit
+
+func dataFromFile(withBundleClass bundleClass: AnyClass, name: String, type: String) -> Data {
+    let filePath = Bundle(for: bundleClass).path(forResource: name, ofType: type)
+
+    guard
+        let path = filePath,
+        let data = try? Data(contentsOf: URL(fileURLWithPath: path))
+    else {
+            fatalError("🔥: file not found or invalid data!")
```
mschuetz72: not sure, but indentions looks 1 tab too much

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
portellaa: 😳😂😇

AlicerceTests/Sources/Persistence/DiskMemoryPersistenceTestCase.swift
portellaa: 😇 ups I forgot

AlicerceTests/Sources/Utils.swift
portellaa: It is. Thanks 😇

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+            guard isDir.boolValue == false else {
+                return completion { throw PersistenceError.other(Error.fileNotCreated) }
+            }
+
+            guard self.fileManager.createFile(atPath: path, contents: data) else {
+                return completion { throw PersistenceError.other(Error.fileNotCreated) }
+            }
+
+            if !fileExists {
```
p4checo: Please be consistent regarding `false` comparisons 😛  (we appear to be using `== false` 😆)

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+        evictOperation.addDependency(writeOperation)
+
+        writeOperationQueue.addOperation(writeOperation)
+        writeOperationQueue.addOperation(evictOperation)
+    }
+
+    private func removeDiskData(for key: Persistence.Key, completion: @escaping CompletionClosure<Void>) {
+        guard diskCacheEnabled else { return completion { throw PersistenceError.other(Error.diskCacheDisabled) } }
+
+        let removeOperation = DiskMemoryBlockOperation(qos: .userInitiated) { [unowned self] in
```
p4checo: The QoS is hardcoded to `.userInitiated`, which is the QoS right after `.userInteractive` (main queue level) 😱. 

Are we sure we need this priority? We could also make it configurable, I guess.

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+
+            return completion { () }
+        }
+
+        writeOperationQueue.addOperation(removeOperation)
+    }
+
+    private func calculateUsedDiskSize() {
+
+        let calculateUsedSizeOperation = DiskMemoryBlockOperation(qos: .userInteractive) { [weak self] in
```
p4checo: Same as above

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+        let removeOperation = DiskMemoryBlockOperation(qos: .userInitiated) { [unowned self] in
+            let path = self.diskPath(for: key)
+
+            do {
+                try self.fileManager.removeItem(atPath: path)
+            } catch let error {
+                return completion { throw PersistenceError.other(Error.failedToRemoveFile(error)) }
+            }
+
+            return completion { () }
```
p4checo: This `return` is not necessary 💄

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+        }
+
+        writeOperationQueue.addOperation(removeOperation)
+    }
+
+    private func calculateUsedDiskSize() {
+
+        let calculateUsedSizeOperation = DiskMemoryBlockOperation(qos: .userInteractive) { [weak self] in
+            guard let `self` = self else { return }
```
p4checo: This will not be allowed in the future, might as well update this now 😢

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+                return []
+        }
+
+        return urls
+    }
+
+    // MARK: - Operations
+
+    private func createEvictOperation() -> DiskMemoryBlockOperation {
+        return DiskMemoryBlockOperation(qos: .background) { [unowned self] in
```
p4checo: hardcoded QoS 😇

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    private func hasDiskCacheDirectory() -> Bool {
+        return fileManager.fileExists(atPath: configuration.path)
+    }
+
+    // FIX: Should this error be propagated to the top level?
+    private func createDiskCacheDirectory() -> Bool {
+        do {
+            try fileManager.createDirectory(atPath: configuration.path, withIntermediateDirectories: true, attributes: nil)
```
p4checo: This line can be broken on the parameters 💄

portellaa:
@mschuetz72 can you review again? already did the changes you requested?

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+            }
+
+            let extraOccupiedDiskSize = self.usedDiskSize - self.configuration.diskLimit
+
+            let urls = self.directoryContents(with: [.contentAccessDateKey, .fileSizeKey])
+
+            typealias FileAccessTimeSizeTuple = (accessTime: TimeInterval, size: UInt64)
+            typealias FilePathAttributesTuple = (path: String, fileAttr: FileAccessTimeSizeTuple)
+
+            let accessTimes: [FilePathAttributesTuple] = urls.map {
```
p4checo: Is `accessTimes` the best name for this variable? 🤔 It seems to store more than that 😝

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+                let resourceValue = $0.resourceValues(forKeys: [.contentAccessDateKey, .fileSizeKey])
+
+                return ($0.path, (resourceValue.contentAccessDate?.timeIntervalSince1970 ?? 0, UInt64(resourceValue.fileSize ?? 0)))
+                }.sorted { $0.fileAttr.accessTime > $1.fileAttr.accessTime }
+
+            var evictSize: UInt64 = 0
+
+            let filesToRemove = accessTimes.filter {
```
p4checo: It somehow feels strange to `filter` the sequence while at the same time incrementing a counter.

Perhaps a more appropriate operator would be the `prefix(while:)`:

```swift
let filesToRemove = accessTimes.prefix { 
    guard evictSize < extraOccupiedDiskSize else { return false }
    evictSize += $0.fileAttr.size
    return true
}
```

The main difference (besides the name) is that this returns an `ArraySlice` instead of an `Array`.

It's just 💄 anyway, so feel free to ignore 😆

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+                    return false
+                }
+
+                evictSize += $0.fileAttr.size
+
+                return true
+            }
+
+
+            filesToRemove.map {$0.path}
```
p4checo: Please add a whitespace between the brackets and the expression 👮

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+            .forEach {
+                guard let _ = try? self.fileManager.removeItem(atPath: $0) else {
+                    assertionFailure("💥 Failed to remove file with path 👉 \($0)")
+
+                    return print("💥 Failed to remove file with path 👉 \($0)")
```
p4checo: We could use our new ✨`Logger`✨  for these messages 😉

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+                }
+            }
+        }
+    }
+}
+
+fileprivate final class DiskMemoryBlockOperation: BlockOperation {
+    
+    required init(qos: QualityOfService? = .default, block: @escaping () -> Swift.Void) {
```
p4checo: Why is the `QualityOfService` optional? It already has a default parameter value set to `.default` (which is already the default for operations),  so perhaps it's a bit irrelevant

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
```diff
@@ -0,0 +1,326 @@
...
+        super.init()
+        
+        addExecutionBlock(block)
+        if let qos = qos { qualityOfService = qos }
+    }
+}
+
+fileprivate extension URL {
```
p4checo: Perhaps this extension is useful for other scenarios? If so we could move it to an :sparkles:`URL.swift`:sparkles: extension 😄

AlicerceTests/Sources/Persistence/DiskMemoryPersistenceTestCase.swift
```diff
@@ -0,0 +1,235 @@
...
+                Thread.sleep(forTimeInterval: 5)
+                if times > 0 {
+                    checkAndRetry(times: times - 1, execute: execute)
+                }
+            }
+
+            execute(retry)
+        }
+
+        waitForExpectations(timeout: 60) {
```
p4checo: Perhaps we could avoid these hardcoded expectations if we made the queues visible as `internal` on the `DiskMemoryPersistence`. 

That way we would simply schedule a block with the `expectation.fulfill()` on the right queue after scheduling the operations.

It would also not be visible to framework users since it would be `internal` 💪

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
portellaa: i prefer to keep this one like this, is an important operation

Alicerce/Sources/Persistence/DiskMemoryPersistence.swift
portellaa: 😮

PR #33, opened 2017-04-19
Improve embedInViewController with generic

portellaa:


Sources/UIKit/UIViewController.swift
```diff
@@ -10,11 +10,18 @@ import UIKit
...
+    public func embedInNavigationController() -> UINavigationController {
+        return embedInNavigationController(custom: UINavigationController.self)
+    }
+    
     /// Embeds `self` in a UINavigationController
     ///
     /// - Returns: UINavigationController with `self` as root
-    public func embedInNavigationController() -> UINavigationController {
-        return UINavigationController(rootViewController: self)
+    public func embedInNavigationController<T: UINavigationController>(custom: T.Type) -> T {
```
p4checo: I would replace `custom:` with `withType type:` 💄

portellaa: makes sense 😉

PR #37, opened 2017-04-20
Removed destination management methods from Logger protocol

### LINKED ISSUE #36: Expose simple protocol for logger
Shouldn't we expose a simple `protocol` with just real logging methods?

Our current implementation is:

```swift
protocol Logger {

    // MARK:- Destination Management

    func register(_ destination: LogDestination)
    func unregister(_ destination: LogDestination)
    func removeAllDestinations()

    // MARK:- Logging

    func verbose(_ message: @autoclosure () -> String,
                 file: StaticString,
                 function: String,
                 line: UInt)

    func debug(_ message: @autoclosure () -> String,
               file: StaticString,
               function: String,
               line: UInt)

    func info(_ message: @autoclosure () -> String,
              file: StaticString,
              function: String,
              line: UInt)

    func warning(_ message: @autoclosure () -> String,
                 file: StaticString,
                 function: String,
                 line: UInt)

    func error(_ message: @autoclosure () -> String,
               file: StaticString,
               function: String,
               line: UInt)
    
    func log(level: Log.Level,
             message: @autoclosure () -> String,
             file: StaticString,
             function: String,
             line: UInt)
}
```

but i think we should have a small protocol only with:

```swift
public protocol Logger {
    func verbose(_ message: @autoclosure () -> String,
                 file: StaticString,
                 function: String,
                 line: UInt)

    func debug(_ message: @autoclosure () -> String,
               file: StaticString,
               function: String,
               line: UInt)

    func info(_ message: @autoclosure () -> String,
              file: StaticString,
              function: String,
              line: UInt)

    func warning(_ message: @autoclosure () -> String,
                 file: StaticString,
                 function: String,
                 line: UInt)

    func error(_ message: @autoclosure () -> String,
               file: StaticString,
               function: String,
               line: UInt)
}
```

when using logger, i can create a global instance of Logger and only access the real logging methods and not `register`, ...

mschuetz72:
Closes #36 🥇

### DIFF
--- a/Sources/Logging/Protocols/Logger.swift
+++ b/Sources/Logging/Protocols/Logger.swift
@@ -12,32 +12,32 @@ protocol Logger {
 
     func verbose(_ message: @autoclosure () -> String,
                  file: StaticString,
-                 function: String,
+                 function: StaticString,
                  line: UInt)
 
     func debug(_ message: @autoclosure () -> String,
                file: StaticString,
-               function: String,
+               function: StaticString,
                line: UInt)
 
     func info(_ message: @autoclosure () -> String,
               file: StaticString,
-              function: String,
+              function: StaticString,
               line: UInt)
 
     func warning(_ message: @autoclosure () -> String,
                  file: StaticString,
-                 function: String,
+                 function: StaticString,
                  line: UInt)
 
     func error(_ message: @autoclosure () -> String,
                file: StaticString,
-               function: String,
+               function: StaticString,
                line: UInt)
     
     func log(level: Log.Level,
              message: @autoclosure () -> String,
              file: StaticString,
-             function: String,
+             function: StaticString,
              line: UInt)
 }
--- a/Sources/Logging/Protocols/Logger.swift
+++ b/Sources/Logging/Protocols/Logger.swift
@@ -8,12 +8,6 @@
 
 protocol Logger {
 
-    // MARK:- Destination Management
-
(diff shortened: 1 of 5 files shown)


p4checo:
@mschuetz72 when addressing an issue, it's common practice to mention it on the PR description, like:

Closes #36 

😉

Sources/Logging/Protocols/Logger.swift
```diff
@@ -8,42 +8,36 @@
...
+                 function: StaticString,
                  line: UInt)
 
     func error(_ message: @autoclosure () -> String,
                file: StaticString,
-               function: String,
+               function: StaticString,
                line: UInt)
     
     func log(level: Log.Level,
```
portellaa: Is this necessary to be public? i think this should be private, no? we have helper methods for all the levels, so no need for this one.

Sources/Logging/Protocols/Logger.swift
```diff
@@ -8,42 +8,36 @@
 
 protocol Logger {
```
portellaa: missing public here

Sources/Logging/Protocols/Logger.swift
mschuetz72: Well, it isn't strictly necessary. I thought it would be better to leave that *generic* to write something into the log open, but you're right. I guess there won't be much cases, if any, this generic method is of any use. Made it private.

PR #39, opened 2017-04-20
Allow tabBarItem helper to customize rendering mode

portellaa:
I think we should provide a way to customize the render mode in the tabBarItem helper so we can toggle if the system should tint the icon or not.

### DIFF
--- a/Sources/UIKit/UIViewController.swift
+++ b/Sources/UIKit/UIViewController.swift
@@ -34,8 +34,11 @@ public extension UIViewController {
     /// - Parameters:
     ///   - selectedImage: UIImage to show when tab is active
     ///   - unselectedImage: UIImage to show when tab is inactive
-    public func tabBarItem(withSelectedImage selectedImage: UIImage?, unselectedImage: UIImage?) {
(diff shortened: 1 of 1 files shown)


Sources/UIKit/UIViewController.swift
```diff
@@ -34,8 +34,11 @@ public extension UIViewController {
     /// - Parameters:
     ///   - selectedImage: UIImage to show when tab is active
     ///   - unselectedImage: UIImage to show when tab is inactive
-    public func tabBarItem(withSelectedImage selectedImage: UIImage?, unselectedImage: UIImage?) {
-        tabBarItem.image = unselectedImage?.withRenderingMode(.alwaysOriginal)
-        tabBarItem.selectedImage = selectedImage?.withRenderingMode(.alwaysOriginal)
+    public func tabBarItem(withSelectedImage selectedImage: UIImage?,
```
mschuetz72: Missing the new parameters in the comments

portellaa: 😂  👍 thanks

PR #48, opened 2017-04-27
Rename `Persistence` to `PersistenceStack`

p4checo:
For consistency’s sake, renamed `Persistence` protocol to
`PersistenceStack`, and created a caseless `enum` `Persistence` to
contain persistence specific types such as `Key` and `Error`.

Moved all CoreData persistence stack code into a separate CoreData
folder.


Sources/Stores/Store.swift
```diff
@@ -0,0 +1,98 @@
+//
+//  Store.swift
+//  mobile_mrp_mrporter
```
portellaa: mobile_mrp_mrporter

portellaa: 🚓

PR #49, opened 2017-04-27
Analytics layer 💣  🙈

portellaa:


Sources/Analytics/Analytics.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+    
+    // MARK: - Public Methods
+    
+    /// Appends a new parameter into the extraParameters.
+    /// Those are appended with tracking item when a track is called.
+    ///
+    /// - Parameter parameters: A dictionary with a String and Any
+    public func add(parameters: Parameters) {
+        if let _ = extraParameters {
```
p4checo: if we are unwrapping here, we can use the unwrapped value and avoid optional chaining 😛

Sources/Analytics/Analytics.swift
```diff
@@ -0,0 +1,96 @@
...
+    /// - Parameter tracker: An AnalyticsTracker object
+    public func add(tracker: AnalyticsTracker) {
+        trackers.append(tracker)
+    }
+}
+
+extension Analytics: AnalyticsTracker {
+    public func track(page: Page) {
+        let trackersCopy = trackers
+        let parameters = merge(parameters: page.parameters, withExtraParameters: extraParameters)
```
p4checo: Why not move this inside the queue too?

Sources/Analytics/Analytics.swift
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+}
+
+private func merge(parameters: Analytics.Parameters?,
+                   withExtraParameters extra: Analytics.Parameters?) -> Analytics.Parameters? {
+    var parameters = parameters
+    
+    extra.then {
+        parameters?.unionInPlace(dictionary: $0)
```
p4checo: If `parameters` is `nil` and `extra` non `nil`, then we will return `nil` when we should return `extra`, right?

Why not convert this logic into a `switch (parameters, extra)`? That way all the cases are explicitly handled 😉 

e.g.:

```swift

switch (parameters, extra) {
case (var parameters?, let extra?): 
    parameters.unionInPlace(extra) // this could be made into a single step if we had something like `union` that returned the result of the union
    return parameters
case (nil, let extra?): 
    return extra
case (let parameters?, nil): 
    return parameters
case (nil, nil)
   return nil
}

```

Sources/Analytics/Configuration.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+extension Analytics {
+    
+    /// Anaylitics configuration.
+    /// Used to configure values in the `Analytics`
+    /// 💣 QueueQoS 👉 The qos to be used by the analytics queue
```
p4checo: Not sure about these emoji in the docs 😇 

Do they appear correctly when Alt+Clicking?

Tests/AlicerceTests/Analytics/AnalyticsTestCase.swift
```diff
@@ -0,0 +1,154 @@
...
+
+        let page = Analytics.Page(name: "trackPage", parameters: nil)
+        let testableTracker = TestableTracker(expectation: expectation, page: page)
+        
+        let analytics = Analytics()
+        analytics.add(tracker: testableTracker)
+        
+        analytics.track(page: page)
+        
+        waitForExpectations(timeout: 30)
```
p4checo: Why not create a constant with this timeout? That way it's easier to change in the future

Tests/AlicerceTests/Analytics/AnalyticsTestCase.swift
```diff
@@ -0,0 +1,154 @@
...
+        let expectation = self.expectation(description: "Track page to the testable tracker")
+        
+        let event = Analytics.Event(name: "trackableEvent", parameters: ["1" : "1"])
+        let tracker = TestableTracker(expectation: expectation, event: event)
+        
+        let analytics = Analytics()
+        analytics.add(tracker: tracker)
+        
+        analytics.track(event: event)
+
```
p4checo: Shouldn't we somehow assert that the parameters match the ones we expect on these tests? Or perhaps I'm missing something 🤔

Tests/AlicerceTests/Analytics/EventTestCase.swift
```diff
@@ -0,0 +1,29 @@
...
+        XCTAssertEqual(event.name, "💣")
+        XCTAssertNil(event.parameters)
+    }
+    
+    func testEvent_WithProvidedNameAndParameters_ItShouldPopulateOnlyTheName() {
+        let event = Analytics.Event(name: "👯", parameters: ["1" : "1"])
+        
+        XCTAssertEqual(event.name, "👯")
+        XCTAssertNotNil(event.parameters)
+        XCTAssertEqual(event.parameters?.count, 1)
```
p4checo: Not exactly a thorough comparison... 😆

Tests/AlicerceTests/Analytics/PageTestCase.swift
```diff
@@ -0,0 +1,29 @@
...
+        XCTAssertEqual(page.name, "🤠")
+        XCTAssertNil(page.parameters)
+    }
+    
+    func testPage_WithProvidedNameAndParameters_ItShouldPopulateOnlyTheName() {
+        let page = Analytics.Page(name: "🔨", parameters: ["1" : "1"])
+        
+        XCTAssertEqual(page.name, "🔨")
+        XCTAssertNotNil(page.parameters)
+        XCTAssertEqual(page.parameters?.count, 1)
```
p4checo: Same as above

Tests/AlicerceTests/Analytics/AnalyticsTestCase.swift
portellaa: should we take that to that level? it looks for me like we are testing not this functionality from others components, not this one.

Sources/Analytics/Configuration.swift
portellaa: yes

p4checo: Ok, at least that 😄 

Anyway, even though I am a fan of emoji 💪 I think that putting them like this on the documentation is probably going too far. But that's my opinion 😇 

Besides one of them being a bomb 😆, they won't be consistent with the rest of the documentation, so I would remove them

Tests/AlicerceTests/Analytics/AnalyticsTestCase.swift
p4checo: How are you testing the parameter merge logic?

portellaa: You know in this case that the count should increase. if you use a different merge strategy, it would stay the same, or anything else. We should test behaviour of this API, not other intern logic.
I should not test internal logic, or neither know internally how this works.

If i send a dictionary with two values and then i send another with 3, they should be 5 if there is no repeated values. if there are repeated values, it can be 3 or 4, we know how many they will be, but depending on the merge strategy it can pick the value from the left dictionary or from the right, we don't know.

PR #51, opened 2017-04-28
Improve `URLSessionNetworkStack`

p4checo:
In order to define `self` as the session's delegate while preserving
dependency injection, the session must be injected via property. This
is because the session's delegate is only defined on its `init`. 🤷‍♂️

The session's delegate could be set to `self` using a lazy var (since
`self` is already defined), but then the session couldn't be injected
for unit testing.

Added a new optional property to the `Network.Configuration` to perform
authentication challenge validations. This is a closure property which
receives the `URLAuthenticationChallenge` from the session’s
`URLSessionDelegate`’s
`urlSession(:didReceive challenge:completionhandler:)` method and
invokes the supplied `completionHandler`.

Until we have a proper server trust validator, this closure can be used
to perform any validations like allowing servers using expired and
self-signed certificates.

### DIFF
--- a/Sources/Network/Network.swift
+++ b/Sources/Network/Network.swift
@@ -13,6 +13,8 @@ public enum Network {
     // MARK: - TypeAlias
 
     public typealias CompletionClosure = (_ inner: () throws -> Data) -> Void
+    public typealias AuthenticationCompletionClosure = (URLSession.AuthChallengeDisposition, URLCredential?) -> Void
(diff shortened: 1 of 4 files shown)


p4checo:
@portella: I agree that this solution is not optimal, hence my comment on the PR 😛:

> Until we have a proper server trust validator, this closure can be used
> to perform any validations like allowing servers using expired and
> self-signed certificates.

We can discuss a nice Validator approach for our Network afterwards. However, given the current time constraints this is good enough, don't you think? 😉  Doing security code in a rush is never a good idea 😆💥

PR #54, opened 2017-05-05
Enhance `View` and create `ViewModelView`

p4checo:
The goal of this family of protocols is to enforce a common base and
lifecycle for all views in a project.

- Modified `View` protocol to have functions `setUpSubviews()` and
`setUpConstraints()`instead of simply `setupLayout`.
- Created `ViewModelView` and `ReusableViewModelView` protocols which
extend `View` and `ReusableView`, respectively. They should
receive a `ViewModel` on `init` and have a `setUpBindings` method. This
enables MVVM views to conform to `ViewModelView` and have a ViewModel
injected (otherwise they couldn’t use `View` with constructor-based
dependency injection).
- Updated CollectionView TableView cells and header/footer views
- Updated folder/project structure and renamed `Protocols` folder to
`View`.
- Added some UT’s for the CollectionView and TableView cells and
headers/footers.
- Improved `UICollectionView` and `UITableView` `ReusableView`
extension methods error messages, and made calls fail with `fatalError`
instead of `assertionFailure`.


Sources/View/ReusableView.swift
```diff
@@ -20,15 +20,19 @@ public extension ReusableView where Self: UITableViewCell {
     init() {
         self.init(style: .default, reuseIdentifier: Self.reuseIdentifier)
 
-        setupLayout()
+        setupSubviews()
```
rbarbosa: As a verb, I would change all methods from `setup[...]` to `setUp[...]`.
What do you think?

p4checo: I have mixed feelings about this. Even though grammatically `setUp` is probably more correct, visually I think I prefer `setup` 😆 

@portellaa, @mschuetz72, @jakewelton opinions?

rbarbosa: In Xcode, when creating tests, the setup method created is `setUp()`.
But whatever people decide is fine for me.

jakewelton: Hmm... personally I too prefer setup, even though setUp is grammatically correct. However, I've just done a quick check to see if Apple use a particular way. `XCTest` uses `setUp`, so maybe we should change to that for consistency?

https://developer.apple.com/reference/xctest/xctestcase/1496262-setup

p4checo: That's a nice point 😉 Didn't remember that 😇

portellaa: 💄

Sources/View/ViewModelView.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by André Pacheco Neves on 05/05/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol ViewModelView: View {
+    associatedtype ViewModel
+
```
portellaa: aren't we missing the property? `var viewModel: ViewModel { get set }`

Sources/View/ReusableView.swift
mschuetz72: Visually I also prefer ```setup```

p4checo: I guess that being both grammatically correct and used by Apple are strong arguments to change it 😛

Let's try and do the **Right Thing ®** 😂

portellaa:
what do you think of this? 

```swift
public protocol ViewModelView: View {
    associatedtype ViewModel

    var viewModel: ViewModel { get set }

    init(viewModel: ViewModel)
}

public extension ViewModelView where Self: UIView {
    init(viewModel: ViewModel) {
        self.init()

        self.viewModel = viewModel
    }
}
```

Sources/View/ViewModelView.swift
p4checo: I left it out on purpose to avoid having to expose it publicly as a property (even though it could be `private(set)`). However we can add that and trust the users will behave and not break encapsulation 😆

p4checo: On the other side, having it like this also allows injecting via property, which may sometimes be useful. 👍

portellaa: i see your point, it makes a lot of sense, but it is developer responsibility. This is a base for something, not a prison 😄 right?

p4checo:
Regarding this PR, while making some unit tests I noticed that the `init`'s are never called. I had already stumbled across this problem in the CoreDataStack, but had forgot about it. 

Even though the code compiles and everything seems to be fine, when executing those `init`'s are never called ( I think the default `NSObject`'s `init` is called instead). Don't know for sure whether it's a bug or a Swift 3 limitation, but the fact is that it currently doesn't work 😢 .

In sum, neither `View` or `ReusableView` protocol extensions' `init()`'s work.

AFAIK, the only way to have this "default behavior" is to define superclasses. Otherwise, we can leave the `init` in the protocol definitions, but it's pointless to try and implement them on protocol extensions.

Not sure if creating these superclasses is worth it, though. Thoughts?

Sources/View/ViewModelView.swift
```diff
@@ -0,0 +1,25 @@
...
+    var viewModel: ViewModel { get set }
+
+    init(viewModel: ViewModel)
+}
+
+public extension ViewModelView where Self: UIView {
+    init(viewModel: ViewModel) {
```
portellaa: this init works, right?

p4checo:
@jakewelton, @mschuetz72, @portellaa, @rbarbosa: Please re-review  since I've changed some things because of the "useless" `init()`'s on the `View` protocol

🙏

PR #55, opened 2017-05-11
Enhance cell helpers

jakewelton:
The current naming of `cell(for indexPath: IndexPath)` could be confused with with `cellForRow(at...`. The call might be expecting our method to return an already dequeued cell, when instead we are dequeueing one. 

This PR prepends the "dequeue" word to the front of the current methods to make this action clear to the caller. It also includes methods to return already dequeued cells of a specific type (essentially calling `cellForRow(at:)` instead of `dequeueReusableCell`)

### DIFF
--- a/Sources/Protocols/ReusableView.swift
+++ b/Sources/Protocols/ReusableView.swift
@@ -47,13 +47,13 @@ public extension UICollectionView {
     }
     
     func cell<T: UICollectionViewCell>(`for` indexPath: IndexPath) -> T
-        where T: ReusableView {
-            guard let cell = cellForItem(at: indexPath) as? T else {
-                assertionFailure("🔥 Cell at \(indexPath) is not of type: `\(T.self)`")
-                return T()
-            }
-            
-            return cell
+    where T: ReusableView {
+        guard let cell = cellForItem(at: indexPath) as? T else {
+            assertionFailure("🔥 Cell at \(indexPath) is not of type: `\(T.self)`")
+            return T()
+        }
+        
+        return cell
     }
 
     func register<T: UICollectionViewCell>(_ cellType: T.Type)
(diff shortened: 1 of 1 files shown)


Sources/Protocols/ReusableView.swift
```diff
@@ -45,6 +45,16 @@ public extension UICollectionView {
 
         return cell
     }
+    
+    func Cell<T: UICollectionViewCell>(`for` indexPath: IndexPath) -> T
```
p4checo: Shouldn't the method name be lowercase?

jakewelton: Yeah it should be. Thanks! Fixed in c8444a957020bcc0b89cab508a2899cce14def6e 😄

Sources/Protocols/ReusableView.swift
```diff
@@ -45,6 +45,16 @@ public extension UICollectionView {
 
         return cell
     }
+    
+    func cell<T: UICollectionViewCell>(`for` indexPath: IndexPath) -> T
+        where T: ReusableView {
```
portellaa: can you fix the indentation here, please? 💄

Sources/Protocols/ReusableView.swift
```diff
@@ -69,13 +79,26 @@ public extension UICollectionView {
 
         return supplementaryView
     }
+    
+    @available(iOS 9, *)
+    func supplementaryView<T: UICollectionReusableView>(forElementKind elementKind: String,
+                                  at indexPath: IndexPath) -> T
+        where T: ReusableView {
```
portellaa: can you fix indentation, please? 💄

Sources/Protocols/ReusableView.swift
```diff
@@ -94,6 +117,36 @@ public extension UITableView {
 
         return view
     }
+    
+    func cell<T: UITableViewCell>(`for` indexPath: IndexPath) -> T
```
portellaa: can you fix the indentation in this methods, please? 💄

Sources/Protocols/ReusableView.swift
jakewelton: Fixed in e9815de6622c0e3fc3706ba68b4867931c309706 :smile:

Sources/Protocols/ReusableView.swift
jakewelton: Fixed in e9815de6622c0e3fc3706ba68b4867931c309706 :smile:

Sources/Protocols/ReusableView.swift
jakewelton: Fixed in e9815de6622c0e3fc3706ba68b4867931c309706 :smile:

p4checo:
@jakewelton Are you still going to add anything to this PR, or can it be merged?

jakewelton:
@p4checo Sorry, yeah I'm finished. I'll merge it 😄

PR #57, opened 2017-05-17
Keyboard Observer

portellaa:
On of the boring thing that iOS developers has to deal, is the keyboard dismissal and setup the location in view where to do it.

This is a simple observer that listens the keyboardDidShow and keyboardDidHide notifications from the `NotificationCenter`, adds a GestureRecognizer into UIWindow and dismisses the keyboard when you press a part of screen where there isn't other touchable elements.

Unfortunately i didn't find yet a way of test it.

The current way of use it, is:
* on you AppDelegate, just create a `private var keyboardObserver: KeyboardObserver!`
* init it in the didFinishLaunching...

If you have a better suggestion, like add a `startObserve()` or a `configureWindow()` 👈 don't like this one, please shout.

🍻

### DIFF
--- a/Sources/Observers/KeyboardObserver.swift
+++ b/Sources/Observers/KeyboardObserver.swift
@@ -6,12 +6,11 @@
 //  Copyright © 2017 Mindera. All rights reserved.
 //
 
-import Foundation
 import UIKit
 
 public final class KeyboardObserver: NSObject {
 
-    fileprivate var keyboardVisible = false
+    fileprivate var isKeyboardVisible = false
 
     private weak var window: UIWindow?
 
@@ -37,20 +36,16 @@ public final class KeyboardObserver: NSObject {
 
     // MARK: - Private Methods
 
-    @objc private func keyboardDidShow(_ notification: Notification) {
-        guard keyboardVisible == false else { return }
-
-        keyboardVisible = true
+    @objc private func keyboardDidShow() {
+        isKeyboardVisible = true
     }
 
-    @objc private func keyboardDidHide(_ notification: Notification) {
-        guard keyboardVisible == true else { return }
-
-        keyboardVisible = false
+    @objc private func keyboardDidHide() {
+        isKeyboardVisible = false
     }
 
     @objc private func didTapView() {
-        guard keyboardVisible == true else { return }
+        guard isKeyboardVisible == true else { return }
 
         window?.endEditing(true)
     }
@@ -58,6 +53,6 @@ public final class KeyboardObserver: NSObject {
 
 extension KeyboardObserver: UIGestureRecognizerDelegate {
     public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
-        return keyboardVisible
+        return isKeyboardVisible
     }
 }
--- a/Sources/Observers/KeyboardObserver.swift
+++ b/Sources/Observers/KeyboardObserver.swift
@@ -1,0 +1,63 @@
+//
+//  KeyboardObserver.swift
+//  Alicerce
+//
+//  Created by Luís Portela on 17/05/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+public final class KeyboardObserver: NSObject {
+
+    fileprivate var keyboardVisible = false
+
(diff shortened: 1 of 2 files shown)


Sources/Observers/KeyboardObserver.swift
```diff
@@ -0,0 +1,63 @@
+//
+//  KeyboardObserver.swift
+//  Alicerce
+//
+//  Created by Luís Portela on 17/05/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
```
p4checo: No need for this `import` 💄

Sources/Observers/KeyboardObserver.swift
```diff
@@ -0,0 +1,63 @@
...
+import UIKit
+
+public final class KeyboardObserver: NSObject {
+
+    fileprivate var keyboardVisible = false
```
p4checo: I would probably prefer `isKeyboardVisible` 💄

Sources/Observers/KeyboardObserver.swift
```diff
@@ -0,0 +1,63 @@
...
+
+        window?.endEditing(true)
+    }
+}
+
+extension KeyboardObserver: UIGestureRecognizerDelegate {
+    public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+        return keyboardVisible
+    }
+}
```
p4checo: Just wondering 🤔 To avoid interfering with other gesture recognizers, shouldn't we implement below delegates and return `false`? 

The methods are:

- `gestureRecognizer(_ gestureRecognizer:, shouldReceive touch:) -> Bool`
- `gestureRecognizer(_ gestureRecognizer:, shouldReceive press:) -> Bool`

Not sure about this, though 😇

Sources/Observers/KeyboardObserver.swift
portellaa: yes i agree, forgot to change this.

Sources/Observers/KeyboardObserver.swift
portellaa: `shouldBegin` returns false when the keyboard is not visible, and so avoiding this gesture to receive touches when is not required.

Sources/Observers/KeyboardObserver.swift
portellaa: nice catch ;)

Sources/Observers/KeyboardObserver.swift
p4checo: Great! 👍

Sources/Observers/KeyboardObserver.swift
```diff
@@ -0,0 +1,63 @@
...
+import Foundation
+import UIKit
+
+public final class KeyboardObserver: NSObject {
+
+    fileprivate var keyboardVisible = false
+
+    private weak var window: UIWindow?
+
+    init(window: UIWindow) {
```
mschuetz72: Not quite sure, but shouldn't this initialiser be ```public```?

Sources/Observers/KeyboardObserver.swift
mschuetz72: Would it make sense to provide delegate methods, like ```keyboardAppeared```and ```keyboardDisappeared``` to provide hooks to use in the client apps? I understand that we've got the NotificationCenter thingy, but whenever I've got to use it I kind of feel dirty... 🦆

Sources/Observers/KeyboardObserver.swift
```diff
@@ -0,0 +1,62 @@
...
+                                               name: .UIKeyboardDidHide,
+                                               object: nil)
+    }
+
+    // MARK: - Private Methods
+
+    @objc private func keyboardDidShow() {
+        guard isKeyboardVisible == false else { return }
+
+        isKeyboardVisible = true
```
jakewelton: I'm sure there is a reason, but I'm curious. Why not just set`isKeyboardVisible` to `true` by default? Rather than doing the `guard`?

portellaa: at the first implementation, i was managing the gesture recognizer, but i removed that, so i can remove that now. nice catch. thanks :)

PR #60, opened 2017-05-25
[WIP] Decouple Store and Resource from Network stack

portellaa:
# What was done 👨‍🔧
## Resource 
* Converted to `protocol`
* Converted `NetworkResource` into protocol
* Added `associatedType` for both type, Remote (the type of data that the remote accepts) and Local (The type of data that client understands)
* Renamed parser into parse (parses remote type to local)
* Added serialize property (serialize local type to remote)
* Added Image network resource to be used with ImageStore

## Store
* Removed logic and properties related with network and persistence
* Created NetworkPersistable store with logic to fetch and persist data using network and persistence stacks
* Create store specific for images (can be useful)

## Parse
* Created ParseError `protocol`
* Moved Error to Parse namespace and extended it from ParseError

## Serialize
* Created Serialize class with image to data implementation
* Created SerializeError `protocol`

# To be done
* Check if it's approved 😅 most important one
* Fix names 😇 second most important
* Fix UnitTests
* Add documentation
* Clean code

Let the 💥 begin


Sources/Network/Parse.swift
```diff
@@ -2,15 +2,17 @@
...
 //  Alicerce
 //
-//  Created by Luís Afonso on 06/04/2017.
+//  Created by Luís Portela on 06/04/2017.
 //  Copyright © 2017 Mindera. All rights reserved.
 //
 
 import UIKit
 
+public protocol ParseError: Swift.Error {}
```
p4checo: What's the advantage of this `protocol`?

Sources/Resource/ImageNetworkResource.swift
```diff
@@ -0,0 +1,37 @@
...
+    public let body: Data? = nil
+
+    public let parse: ResourceMapClosure<Remote, Local> = Parse.image
+    public let serialize: ResourceMapClosure<Local, Remote> = Serialize.imageAsPNGData
+
+    public init(url: URL,
+                path: String? = nil,
+                method: HTTP.Method = .GET,
+                headers: HTTP.Headers? = nil,
+                query: HTTP.Query? = nil) {
```
p4checo: Why isn't `body` on the `init` as well? 🤔

Sources/Resource/Resource.swift
```diff
@@ -8,62 +8,12 @@
...
-    }
-
-    private func buildQueryItems() -> [URLQueryItem]? {
-        guard let query = query, query.isEmpty == false else {
-            return nil
-        }
-
-        return query.map { URLQueryItem(name: $0, value: $1) }
-    }
+    var parse: ResourceMapClosure<Remote, Local> { get }
```
p4checo: For consistency, perhaps I would call this `deserialize`

Sources/Resource/NetworkResource.swift
```diff
@@ -8,12 +8,47 @@
 
 import Foundation
 
-public typealias ResourceParseClosure<T> = (Data) throws -> T
+public protocol NetworkResource: Resource {
```
p4checo: I would perhaps break this protocol into simpler, more specific ones:

```swift

public protocol NetworkResource: Resource {

    var method: HTTP.Method { get }
    var headers: HTTP.Headers? { get }
    var query: HTTP.Query? { get }
    var body: Data? { get }

    func toRequest() -> URLRequest
}

public protocol StaticNetworkResource: NetworkResource {

    var url: URL { get }
}

public extension StaticNetworkResource {

    func toRequest() -> URLRequest {
        var urlRequest = URLRequest(url: url)

        urlRequest.allHTTPHeaderFields = headers
        urlRequest.httpBody = body
        urlRequest.httpMethod = method.rawValue

        return urlRequest
    }
}

public protocol RelativeNetworkResource: NetworkResource {

    static var baseURL: URL { get }

    var path: String { get }
}

public extension RelativeNetworkResource {

    func toRequest() -> URLRequest {
        var url = Self.baseURL

        if var components = URLComponents(url: url, resolvingAgainstBaseURL: false) {
            components.queryItems = buildQueryItems()
            components.path = components.path
                .appending(path)
                .replacingOccurrences(of: "//", with: "/")

            components.url.then {
                url = $0
            }
        }

        var urlRequest = URLRequest(url: url)

        urlRequest.allHTTPHeaderFields = headers
        urlRequest.httpBody = body
        urlRequest.httpMethod = method.rawValue

        return urlRequest
    }

    private func buildQueryItems() -> [URLQueryItem]? {
        guard let query = query, query.isEmpty == false else {
            return nil
        }

        return query.map { URLQueryItem(name: $0, value: $1) }
    }
}

```

This avoids having both `url` and `path` (and simultaneously optional), and makes the `toRequest` API more transparent.

The downside is that we might have to create more types (e.g. `StaticImageNetworkResource` and `RelativeImageNetworkResource`). However I think it may be worth it.

What do you think? 😝

Sources/Stores/DataStore.swift
```diff
@@ -8,14 +8,14 @@
...
-    
-    public let networkStack: NetworkStack
-    public let persistenceStack: P
-    
-    public init(networkStack: NetworkStack, persistenceStack: P) {
-        self.networkStack = networkStack
-        self.persistenceStack = persistenceStack
-    }
-}
+//public final class DataStore<U, P: PersistenceStack>: Store {
```
p4checo: Do we want to keep this around? 😆

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -0,0 +1,122 @@
...
+    private let persistenceStack: PersistenceStack
+
+    public init(networkStack: NetworkStack, persistenceStack: PersistenceStack) {
+        self.networkStack = networkStack
+        self.persistenceStack = persistenceStack
+    }
+
+    @discardableResult
+    public func fetch<Resource: NetworkResource & PersistableResource>(resource: Resource,
+                                                                _ completion: @escaping StoreCompletionClosure<Resource.Local>)
```
p4checo: bad alignment 👮 

I also think that this line is probably over 120 chars 😇

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -0,0 +1,122 @@
...
+                    let value = try resource.parse(data)
+
+                    guard cancelable.isCancelled == false else { return completion(nil, Error.cancelled, false) }
+
+                    // update persistence with new value
+                    self.persist(data, for: resource)
+
+                    completion(value, nil, false)
+                } catch let Network.Error.url(error as NSError) where error.domain == NSURLErrorDomain
+                    && error.code == NSURLErrorCancelled {
```
p4checo: this condition could be aligned with the other one 💄

Sources/Network/Serialize.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Serialize.swift
+//  Alicerce
+//
+//  Created by Luís Portela on 24/05/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol SerializeError: Swift.Error {}
```
p4checo: Same as above

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -0,0 +1,122 @@
...
+                // TODO: remove from persistence?
+                return fetch(resource: resource)
+            }
+        }
+
+        return cancelable
+    }
+
+    private func getPersistedData<Resource: PersistableResource>(for resource: Resource,
+                                  completion: @escaping (Data?) -> ()) {
```
p4checo: bad alignment 🚔

Sources/Stores/Store.swift
```diff
@@ -2,123 +2,17 @@
...
-    private func persist<Resource: PersistableResource>(_ data: Data, for resource: Resource) {
-        persistenceStack.setObject(data, for: resource.persistenceKey) { (inner: () throws -> Void) -> Void in
-            do {
-                try inner()
-            } catch {
-                print("⚠️: Failed to persist value for resource \"\(resource)\"! Error:\(error)")
-            }
-        }
-    }
+//    func fetch<R>(resource: R, _ completion: @escaping StoreCompletionClosure<R.T>) -> Alicerce.Cancelable
```
p4checo: Shouldn't these lines be uncommented?

Sources/Network/Parse.swift
portellaa: honestly, currently i don't know 🤔 as far as i can see it was only to have a base ParseError, but doesn't makes sense now. i'm not even using it in the app the i redefine the errors.

should remove it?

Sources/Resource/ImageNetworkResource.swift
portellaa: hum, the body should be populated using the serialize function, right?

Sources/Resource/Resource.swift
portellaa: but i don't like the name 😢 but ok, i will change

Sources/Network/Parse.swift
p4checo: We could have some extra type safety by forcing all errors coming out of the parse step to conform to this protocol, but then it doesn't model any specific behavior so I don't see a clear advantage.

Sources/Resource/NetworkResource.swift
portellaa: How do we inject the url in the `RelativeNetworkResource` ? From my understanding, it would be inject by the network stack in the resource so it can build the `URLRequest`

Sources/Stores/DataStore.swift
portellaa: 💨

Sources/Resource/ImageNetworkResource.swift
p4checo: I guess you're right 😅

Sources/Stores/NetworkPersistableStore.swift
portellaa: fixed

Sources/Stores/NetworkPersistableStore.swift
portellaa: fixed

Sources/Stores/NetworkPersistableStore.swift
portellaa: fixed

Sources/Stores/Store.swift
portellaa: 😇

Sources/Stores/DataStore.swift
```diff
@@ -8,16 +8,28 @@
...
-    public let persistenceStack: P
-    public let metricsConfiguration: StoreMetricsConfiguration<T>?
-    
-    public init(networkStack: NetworkStack, persistenceStack: P, metricsConfiguration: StoreMetricsConfiguration<T>?) {
-        self.networkStack = networkStack
-        self.persistenceStack = persistenceStack
-        self.metricsConfiguration = metricsConfiguration
-    }
-}
+//public final class DataStore<U, P: PersistenceStack>: Store {
```
p4checo: 🙈

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -90,21 +93,23 @@ public extension Network {
         }
 
         @discardableResult
-        public func fetch<R: NetworkResource>(resource: R,
-                                              completion: @escaping Network.CompletionClosure) -> Cancelable {
+        public func fetch<R>(resource: R,
+                             _ completion: @escaping Network.CompletionClosure<R.Remote>)
```
p4checo: same as above

Sources/Network/NetworkStack.swift
```diff
@@ -9,5 +9,8 @@
 import Foundation
 
 public protocol NetworkStack {
-    func fetch<R: NetworkResource>(resource: R, completion: @escaping Network.CompletionClosure) -> Cancelable
+    associatedtype Remote
+
+    func fetch<R: NetworkResource>(resource: R, _ completion: @escaping Network.CompletionClosure<R.Remote>)
```
p4checo: I think we can remove these anonymous parameters, because we can use trailing closures even if they are named, and when using variables it isn't so weird 😉

Sources/Resource/NetworkResource.swift
```diff
@@ -8,15 +8,82 @@
...
+            components.url.then { newUrl = $0 }
+        }
+
+        return buildRequest(for: newUrl, method: method, headers: headers, body: body)
+    }
+}
+
+private func build(queryItems: HTTP.Query?) -> [URLQueryItem]? {
+    guard let queryItems = queryItems, queryItems.isEmpty == false else {
+        return nil
```
p4checo: this can be on the same line 💄

portellaa:
Here it goes 👋 💣 🔥 🚒

p4checo:
Weeee 🚀 
Excellent work @portellaa 👏😄  
This really is a great architectural improvement 😍

PR #71, opened 2017-07-05
Requests Network authenticator 🔑

portellaa:
Created `Authenticator` protocol to support requests authentication and retry based on status code

### DIFF
--- a/Sources/Network/NetworkAuthenticator.swift
+++ b/Sources/Network/NetworkAuthenticator.swift
@@ -1,0 +1,29 @@
+//
+//  Authenticator.swift
+//  Alicerce
+//
+//  Created by Luís Portela on 03/07/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void)
+
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool
+}
+
+public extension NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void) {
+        completion(request)
+    }
+
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool {
+        guard error == nil else { return false }
+
+        guard let response = response else { return false }
+
+        return response.statusCode == 403
+    }
+}
--- a/Sources/Network/NetworkAuthenticator.swift
+++ b/Sources/Network/NetworkAuthenticator.swift
@@ -9,7 +9,9 @@
 import Foundation
 
 public protocol NetworkAuthenticator {
-    func authenticate(request: URLRequest, _ performRequest: @escaping (URLRequest) -> Cancelable) -> Cancelable
+    typealias PerformRequestClosure = (_ inner: () throws -> URLRequest) -> Cancelable
+
+    func authenticate(request: URLRequest, _ performRequest: @escaping PerformRequestClosure) -> Cancelable
 
     func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool
 }
--- a/Sources/Network/NetworkAuthenticator.swift
+++ b/Sources/Network/NetworkAuthenticator.swift
@@ -1,5 +1,5 @@
 //
-//  Authenticator.swift
+//  NetworkAuthenticator.swift
 //  Alicerce
 //
 //  Created by Luís Portela on 03/07/2017.
@@ -9,21 +9,7 @@
 import Foundation
 
 public protocol NetworkAuthenticator {
-    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void)
+    func authenticate(request: URLRequest, _ performRequest: @escaping (URLRequest) -> Cancelable) -> Cancelable
 
     func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool
-}
-
-public extension NetworkAuthenticator {
-    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void) {
-        completion(request)
-    }
-
-    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool {
-        guard error == nil else { return false }
-
-        guard let response = response else { return false }
-
-        return response.statusCode == 403
-    }
 }
--- a/Sources/Network/URLSessionNetworkStack.swift
+++ b/Sources/Network/URLSessionNetworkStack.swift
@@ -15,7 +15,11 @@ public extension Network {
         private struct CancelableTask: Cancelable {
             private weak var task: URLSessionTask?
 
-            init(task: URLSessionTask) {
+            init(task: URLSessionTask? = nil) {
+                self.task = task
+            }
+
+            mutating func set(task: URLSessionTask) {
                 self.task = task
             }
 
@@ -24,10 +28,23 @@ public extension Network {
             }
         }
 
-        private typealias URLSessionDataTaskClosure = (Data?, URLResponse?, Swift.Error?) -> Void
+        private typealias RequestConfig<R: NetworkResource> = (
+            resource: R,
+            cancelableTask: CancelableTask,
+            completion: Network.CompletionClosure
+        )
+
+//        private struct RequestConfig<R: NetworkResource> {
+//            let resource: R
+//            var cancelableTask: CancelableTask
+//            let completion: Network.CompletionClosure
+//        }
+
+        public typealias URLSessionDataTaskClosure = (Data?, URLResponse?, Swift.Error?) -> Void
 
         private let baseURL: URL
         private let authenticationChallengeValidator: AuthenticationChallengeValidatorClosure?
+        private let authenticator: NetworkAuthenticator?
 
         public var session: URLSession? {
             // In order to define `self` as the session's delegate while preserving dependency injection, the session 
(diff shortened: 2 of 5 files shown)


Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -24,10 +28,23 @@ public extension Network {
...
         }
 
-        private typealias URLSessionDataTaskClosure = (Data?, URLResponse?, Swift.Error?) -> Void
+        private typealias RequestConfig<R: NetworkResource> = (
+            resource: R,
+            cancelableTask: CancelableTask,
+            completion: Network.CompletionClosure
+        )
+
+//        private struct RequestConfig<R: NetworkResource> {
```
jakewelton: Should this still be here? 😄

portellaa: No 😂

Sources/Network/NetworkAuthenticator.swift
```diff
@@ -0,0 +1,29 @@
...
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool
+}
+
+public extension NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void) {
+        completion(request)
+    }
+
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool {
+        guard error == nil else { return false }
```
mschuetz72: Would it make sense to merge the two `guard` statements to one single guard. Anyway, just 💄

Sources/Network/NetworkAuthenticator.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Created by Luís Portela on 03/07/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void)
+
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool
```
p4checo: Don't know if it makes sense, but this closure parameters are somehow "tied" to `URLSession` completion parameters. So it made me think if this protocol could have an `associatedtype` with the completion "type", that would be used for the retry closure.

What do you say? 😇

Sources/Network/NetworkAuthenticator.swift
```diff
@@ -0,0 +1,29 @@
...
+
+public extension NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void) {
+        completion(request)
+    }
+
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool {
+        guard error == nil else { return false }
+
+        guard let response = response else { return false }
```
p4checo: we could perhaps drop this `guard` and simply `return response?.statusCode == 403` 💄

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -15,7 +15,11 @@ public extension Network {
         private struct CancelableTask: Cancelable {
             private weak var task: URLSessionTask?
 
-            init(task: URLSessionTask) {
+            init(task: URLSessionTask? = nil) {
```
p4checo: this doesn't make sense to me, why do we need this? 🤔

Sources/Network/NetworkAuthenticator.swift
portellaa: It does. changed thanks

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -46,31 +63,28 @@ public extension Network {
...
         @discardableResult
         public func fetch<R: NetworkResource>(resource: R,
                                              _ completion: @escaping Network.CompletionClosure) -> Cancelable {
-            guard let session = session else {
-                fatalError("🔥: session is `nil`! Forgot to 💉?")
-            }
+            let cancelableTask = CancelableTask()
 
-            let request = resource.toRequest(withBaseURL: baseURL)
+            performRequest(with: (resource: resource, cancelableTask: cancelableTask, completion: completion))
```
p4checo: why doesn't ` performRequest` simply return the ` CancelableTask` so that we don't need the above `init`?

Sources/Network/NetworkAuthenticator.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Alicerce
+//
+//  Created by Luís Portela on 03/07/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void)
```
p4checo: 1. Is `completion` the right name for this closure? Isn't it the request performing logic?
2. Since this essentially performs N requests, it should return a `Cancelable`, right?

Sources/Network/NetworkAuthenticator.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    func shouldRetry(with data: Data?, response: HTTPURLResponse?, error: Swift.Error?) -> Bool
+}
+
+public extension NetworkAuthenticator {
+    func authenticate(request: URLRequest, _ completion: @escaping (URLRequest) -> Void) {
```
p4checo: I don't think this extension method brings much value, because it won't authenticate anything 😢

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -89,24 +103,62 @@ public extension Network {
...
+        private func performRequest<R: NetworkResource>(with config: RequestConfig<R>) {
+            guard let session = session else {
+                fatalError("🔥: session is `nil`! Forgot to 💉?")
+            }
+
+            var (resource, cancelableTask, _) = config
+
+            let request = resource.toRequest(withBaseURL: baseURL)
+
+            authenticate(request: request) { [weak self] request in
```
p4checo: `authenticate` should somehow return a `Cancelable`, I think

Sources/Network/NetworkAuthenticator.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Authenticator.swift
+//  Alicerce
+//
+//  Created by Luís Portela on 03/07/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol NetworkAuthenticator {
```
p4checo: shouldn't we have some sort of retry limit?

Sources/Network/NetworkAuthenticator.swift
portellaa: 1. about the naming, for me it seems ok, i'm calling the function to `authenticate` the request and then it returns the completion when it's ready, so it seems like a completion closure, however shout your name suggestion if you have one.
2. When you mean requests, are those network requests? Not necessarily, they can be something that takes more time.

Sources/Network/NetworkAuthenticator.swift
portellaa: No.

Why use and abuse of `associatedType` and add an unnecessary complexity. This is called `NetworkAuthenticator`, if this was called `EverythingAuthenticator`, i would agree with that.
But please explain me the advantage of add that complexity.

Sources/Network/URLSessionNetworkStack.swift
portellaa: And do what with that one? The task will be created only after the authenticate finishes and `fetch` has to return to the caller a `Cancelable`

Sources/Network/NetworkAuthenticator.swift
p4checo: 1. I see your point. 
So the "happy path" of `authenticate` is supposed to simply perform the request, correct? 
The "sad path" is when the request fails and then the authentication is made automatically and the request is retried, yes? 
In conclusion, the `completion` should simply signal when either of these cases end. So I'm not understanding how we are sending request creation/validation logic below inside it. (`URLSessionNetworkStack.swift:115`)

2. When I mean requests I mean that the authenticator might need to make more than the one request we give it. What other time consuming operations will it make?

Sources/Network/NetworkAuthenticator.swift
p4checo: Did I say something wrong? I don't think this attitude is constructive. 

If you *read* what I wrote, you will be able to find the exact argument I'm using. 

You think it's unnecessary complexity, fine, just say so. Geez

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -10,36 +10,48 @@ import Foundation
...
+    public final class CancelableBag: Cancelable {
+        private lazy var cancelables: [Cancelable] = []
 
-            public func cancel() {
-                task?.cancel()
-            }
+        public init() {}
+
+        deinit {
+            cancel()
```
p4checo: not sure about this. I might want to discard the `Cancelable`, which may cause it do be `deinit`ed when I simply want the task to complete

portellaa: yes, makes sense. thanks for notice 😉

portellaa: done

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -164,10 +154,14 @@ public extension Network {
...
 
-                if strongSelf.shouldRetry(with: data, response: httpResponse, error: error) {
-                    return cancelableBag.add(cancelable: strongSelf.perform(request: originalRequest,
-                                                                            apiErrorParser: apiErrorParser,
-                                                                            completion))
+                if let authenticator = strongSelf.authenticator,
+                    authenticator.shouldRetry(with: data, response: httpResponse, error: error) {
+
+                    let cancelableRetry = strongSelf.networkAuthenticator(authenticator,
+                                                               perform: originalRequest,
```
p4checo: bad alignment 🚓

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -164,10 +154,14 @@ public extension Network {
...
                 }
 
-                if strongSelf.shouldRetry(with: data, response: httpResponse, error: error) {
-                    return cancelableBag.add(cancelable: strongSelf.perform(request: originalRequest,
-                                                                            apiErrorParser: apiErrorParser,
-                                                                            completion))
+                if let authenticator = strongSelf.authenticator,
+                    authenticator.shouldRetry(with: data, response: httpResponse, error: error) {
+
+                    let cancelableRetry = strongSelf.networkAuthenticator(authenticator,
```
p4checo: `retryCancelable`  💄

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -95,14 +95,8 @@ public extension Network {
...
-            return authenticator.authenticate(request: request) { [weak self] authenticatedRequest -> Cancelable in
-
-                guard let strongSelf = self else { return NoCancelable() }
-
-                return strongSelf.perform(request: authenticatedRequest,
-                                          apiErrorParser: resource.apiErrorParser,
-                                          completion)
-            }
+            return networkAuthenticator(authenticator,
+                                        perform: request, apiErrorParser: resource.apiErrorParser, completion)
```
p4checo: bad alignment 👮

portellaa: 🚒

Sources/Network/URLSessionNetworkStack.swift
portellaa: 🚶

Sources/Network/URLSessionNetworkStack.swift
portellaa: 🚒

PR #72, opened 2017-07-07
Improve JSON parser API's and add optional parse

p4checo:
On some scenarios, using type inference on the result type alone is
cumbersome, e.g. makes chaining parse calls impossible, and using it on
`if` or `guard` statements is also ugly.

Another common used approach (also used in Swift 4's new `Codable`) is
to receive the type as a parameter of the API itself. This makes the
API more flexible and solves the above issues.

To address these limitations, two new API's were added that work over 
specified types instead of inferred ones.

Additionally, some models have optional attributes, which can appear or
not in the JSON. This was not properly addressed in the current
implementation, so a new API `parseOptionalAttribute` was created to
specifically handle these cases. When the attribute key doesn't exist,
an attempt is made to parse a custom API error, and if it doesn't exist,
 `nil` is returned. If, on the other side the key does exist,
validation is made like on `parseAttribute`.

### DIFF
--- a/Sources/Network/JSON.swift
+++ b/Sources/Network/JSON.swift
@@ -61,18 +61,22 @@ public enum JSON {
         return try parse(from: data)
     }
 
-    /// Parse an attribute of type `T` with the given attribute on the given JSON dictionary, and optionally validate 
+    // MARK: - Specified Type
+
+    /// Parse an attribute of type `T` with the given key on the given JSON dictionary, and optionally validate
     /// the value using a `where` predicate closure. Additionally, an optional `parseAPIError` closure can be supplied
     /// so that if a parsing step fails an attempt is made to extract a domain specific error and throw it.
     ///
     /// - Parameters:
-    ///   - attribute: The JSON attribute key to parse.
+    ///   - type: The type of the attribute to parse.
+    ///   - key: The JSON attribute key to parse.
     ///   - json: The JSON dictionary.
     ///   - predicate: The validation predicate.
     ///   - parseAPIError: The API error parsing closure.
-    /// - Returns: The attribute of type `T` asociated to the given key.
+    /// - Returns: The value of type `T` associated to the given attribute's key.
     /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
-    public static func parseAttribute<T>(_ key: JSON.AttributeKey,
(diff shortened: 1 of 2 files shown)


Sources/Network/JSON.swift
```diff
@@ -61,37 +61,55 @@ public enum JSON {
...
+    ///   - type: The type of the attribute to parse.
+    ///   - key: The JSON attribute key to parse.
     ///   - json: The JSON dictionary.
     ///   - predicate: The validation predicate.
     ///   - parseAPIError: The API error parsing closure.
-    /// - Returns: The attribute of type `T` asociated to the given key.
+    /// - Returns: The value of type `T` associated to the given attribute's key.
     /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
-    public static func parseAttribute<T>(_ key: JSON.AttributeKey,
+    public static func parseAttribute<T>(_ type: T.Type,
```
mschuetz72: Wouldn't it make sense to maintain the older version as well (backward compatibility & maintain possibility to take advantage of type inference)?

p4checo: Well, I honestly don't think it's necessary to think about backwards compatibility at this point of the framework's life. 

So far we have consciously avoided versioning the framework precisely to give us complete flexibility over the API's and not have to think about those types of issues until we feel it has reached a certain level of maturity.

In the future I believe we intend to follow semantic versioning and then we will have to concern ourselves with these issues whenever we "break" or deprecate an API.

As for taking advantage of the type inference, do you really think it's worth it?

portellaa: i think this could make sense, however we don't have a release yet to have to maintain backward compatibility.
but we could leave this, it doesn't hurts i think 😄 
But it's up to the developer to decide what she wants to do 😉

mschuetz72: You're right, backwards compatibility may not be an issue here, yet, however requiring the developer to specify the type whereas it could be inferred seems a little *offensive* to me. But, of cause, it's just my personal opinion :)

p4checo: Well, it certainly doesn't hurt, but will be 2 more API's to support in the future 😆 

Do you guys think they bring that much value?

p4checo: I'll update the PR to include both sets of API's since in some simpler cases it might be easier to use the type inferred API's 😉

PR #74, opened 2017-07-21
UIColor to hex string extension method

mschuetz72:
Little UIColor helper method to convert any UIColor to a (HTML)-hex string representation, with or without alpha component.

### DIFF
--- a/Sources/UIKit/UIColor.swift
+++ b/Sources/UIKit/UIColor.swift
@@ -41,20 +41,20 @@ public extension UIColor {
 
         var components = UIColor.components(fromHex6: 0)
         getRed(&components.red, green: &components.green, blue: &components.blue, alpha: &components.alpha)
-        let rgb: Int = (Int)(components.red*255)<<16
-            | (Int)(components.green*255)<<8
-            | (Int)(components.blue*255)<<0
+        let rgb: Int = (Int)(components.red * UIColor.divisor) << 16
+            | (Int)(components.green * UIColor.divisor) << 8
+            | (Int)(components.blue * UIColor.divisor) << 0
         return String(format:"#%06x", rgb)
     }
 
     var hexStringWithAlpha: String {
 
-        var components = UIColor.components(fromHex6: 0)
+        var components = UIColor.components(fromHex8: 0)
         getRed(&components.red, green: &components.green, blue: &components.blue, alpha: &components.alpha)
-        let argb: Int = (Int)(components.alpha * 255)<<24
-            | (Int)(components.red*255)<<16
-            | (Int)(components.green*255)<<8
-            | (Int)(components.blue*255)<<0
+        let argb: Int = (Int)(components.alpha * UIColor.divisor) << 24
+            | (Int)(components.red * UIColor.divisor) << 16
+            | (Int)(components.green * UIColor.divisor) << 8
+            | (Int)(components.blue * UIColor.divisor) << 0
         return String(format:"#%08x", argb)
     }
 
@@ -69,10 +69,11 @@ public extension UIColor {
     }
 
     private static func components(fromHex8 hex: UInt32) -> Components {
-        let red = CGFloat((hex & 0xFF000000) >> 24) / UIColor.divisor
-        let green = CGFloat((hex & 0x00FF0000) >> 16) / UIColor.divisor
-        let blue = CGFloat((hex & 0x0000FF00) >> 8) / UIColor.divisor
-        let alpha = CGFloat(hex & 0x000000FF) / UIColor.divisor
+
+        let alpha = CGFloat((hex & 0xFF000000) >> 24) / UIColor.divisor
+        let red = CGFloat((hex & 0x00FF0000) >> 16) / UIColor.divisor
+        let green = CGFloat((hex & 0x0000FF00) >> 8) / UIColor.divisor
+        let blue = CGFloat(hex & 0x000000FF) / UIColor.divisor
         
         return (red, green, blue, alpha)
     }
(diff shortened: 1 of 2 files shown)


Sources/UIKit/UIColor.swift
```diff
@@ -37,6 +37,27 @@ public extension UIColor {
...
+            | (Int)(components.green*255)<<8
+            | (Int)(components.blue*255)<<0
+        return String(format:"#%06x", rgb)
+    }
+
+    var hexStringWithAlpha: String {
+
+        var components = UIColor.components(fromHex6: 0)
+        getRed(&components.red, green: &components.green, blue: &components.blue, alpha: &components.alpha)
+        let argb: Int = (Int)(components.alpha * 255)<<24
```
p4checo: I'm a bit confused right now. It seems we are using `#RRGGBBAA` when **encoding** a color to hex, but using `#AARRGGBB` when **decoding** from hex 🤔 

Still, I can't find any type of color standard for hex representation with alpha channel...

We should definitely set this straight and settle on one format, and also properly document the API's, imo.

cc  @portellaa

Sources/UIKit/UIColor.swift
```diff
@@ -37,6 +37,27 @@ public extension UIColor {
         self.init(red: components.red, green: components.green, blue: components.blue, alpha: components.alpha)
     }
 
+    var hexString: String {
+
+        var components = UIColor.components(fromHex6: 0)
+        getRed(&components.red, green: &components.green, blue: &components.blue, alpha: &components.alpha)
+        let rgb: Int = (Int)(components.red*255)<<16
```
p4checo: Additionally, we could use `UIColor.divisor` instead of `255` 😉

Tests/AlicerceTests/UIKit/UIColorTestCase.swift
```diff
@@ -38,4 +38,26 @@ final class UIColorTestCase: XCTestCase {
...
+
+        let hexOpaque = UIColor(white: 1.0, alpha: 1.0).hexStringWithAlpha
+        let hexSemi = UIColor(white: 1.0, alpha: 0.5).hexStringWithAlpha
+        let hexTransparent = UIColor(white: 1.0, alpha: 0.0).hexStringWithAlpha
+
+        XCTAssertEqual(hexOpaque, "#ffffffff")
+        XCTAssertEqual(hexSemi, "#7fffffff")
+        XCTAssertEqual(hexTransparent, "#00ffffff")
+    }
 }
```
p4checo: following the above comments regarding the hex format, I would suggest we add some tests encoding  <-> decoding using our own API's and validating that the colors effectively match 😄

Sources/UIKit/UIColor.swift
p4checo: Ok, so it seems that the `ARGB` format is the one that is more often expressed in 8 hexadecimal digits, following the pattern `#AARRGGBB`

The RGBA format can also be represented this way, but seems less frequent.

Source: https://en.wikipedia.org/wiki/RGBA_color_space

Sources/UIKit/UIColor.swift
```diff
@@ -37,6 +37,27 @@ public extension UIColor {
...
+            | (Int)(components.green*255)<<8
+            | (Int)(components.blue*255)<<0
+        return String(format:"#%06x", rgb)
+    }
+
+    var hexStringWithAlpha: String {
+
+        var components = UIColor.components(fromHex6: 0)
```
mschuetz72: Actually it's just an initialisation to 8, so ```fromHex8``` does the same as ```fromHex6```.  I changed it anyhow.

Sources/UIKit/UIColor.swift
portellaa: I agree with decide a format and use that format all the way. at the time i did the hex to rgb components, i based my self in rgba, but it seems that this was changed.

Honestly, argb makes more sense to me rather then rbga, so change let's change it 😄 

@mschuetz72 if you prefer, i can open a PR with the change to my bug, if you don't want to worry with that 😉

mschuetz72: The format that I know and used, I believe is ```#AARRGGBB```. Not sure if there was a specific reason or use case why we've got ```#RRGGBBAA``` for initialising ```UIColor``` objects.

mschuetz72: @portellaa @p4checo if it's OK I'll make that change in the same PR (this one)

Sources/UIKit/UIColor.swift
```diff
@@ -37,6 +37,27 @@ public extension UIColor {
         self.init(red: components.red, green: components.green, blue: components.blue, alpha: components.alpha)
     }
 
+    var hexString: String {
+
+        var components = UIColor.components(fromHex6: 0)
+        getRed(&components.red, green: &components.green, blue: &components.blue, alpha: &components.alpha)
+        let rgb: Int = (Int)(components.red * UIColor.divisor) << 16
```
jakewelton: Do we need the type inference here? I only ask because the Mindera Swift style guide says to use compiler inferred types where possible :smile:

Sources/UIKit/UIColor.swift
```diff
@@ -37,6 +37,27 @@ public extension UIColor {
...
+            | (Int)(components.green * UIColor.divisor) << 8
+            | (Int)(components.blue * UIColor.divisor) << 0
+        return String(format:"#%06x", rgb)
+    }
+
+    var hexStringWithAlpha: String {
+
+        var components = UIColor.components(fromHex8: 0)
+        getRed(&components.red, green: &components.green, blue: &components.blue, alpha: &components.alpha)
+        let argb: Int = (Int)(components.alpha * UIColor.divisor) << 24
```
jakewelton: Same question here regarding type inference 😄

PR #75, opened 2017-07-31
Add completion closure to `RouteHandler` 📥

p4checo:
For some use cases, it is handy to know when a route has effectively
been handled by its `RouteHandler`, instead of losing track of the
execution flow and simply delegating the handling logic.

To address this, a new optional completion closure parameter of type
`((T) -> Void)?` was added to `RouteHandler`'s and `TreeRouter`'s
`handle` methods. Consequently, `RouteHandler` gained a new associated
type `T`, and `TreeRouter` a new generic constraint.

Using a generic type seemed like a nice balance because not only is
completion signaled to the `route` method caller, but the flexibility
to return a custom payload is also provided. This allows the Routing
infrastructure to be used in even more scenarios.

Fixed some synchronization issues on Logging tests and 
`StoreTestCase` which were breaking Travis builds.

### DIFF
--- a/Sources/DeepLinking/Router.swift
+++ b/Sources/DeepLinking/Router.swift
@@ -9,7 +9,9 @@
 import UIKit
 
 public protocol RouteHandler {
-    func handle(route: URL, parameters: [String : String], queryItems: [URLQueryItem])
+    associatedtype T
+
+    func handle(route: URL, parameters: [String : String], queryItems: [URLQueryItem], completion: ((T) -> Void)?)
 }
 
 public protocol Router {
@@ -18,10 +20,10 @@ public protocol Router {
     func register(_ route: URL, handler: Handler) throws
     func unregister(_ route: URL) throws -> Handler
 
-    func route(_ route: URL) throws
+    func route(_ route: URL, completion: ((Handler.T) -> Void)?) throws
 }
 
-public final class TreeRouter<Handler: RouteHandler>: Router {
+public final class TreeRouter<Handler: RouteHandler, T>: Router where Handler.T == T {
 
     public typealias Match = Route.Tree<Handler>.Match
 
@@ -100,7 +102,7 @@ public final class TreeRouter<Handler: RouteHandler>: Router {
         }
     }
 
-    public func route(_ route: URL) throws {
+    public func route(_ route: URL, completion: ((T) -> Void)? = nil) throws {
         let (scheme, routeComponents, queryItems) = try parseRoute(route)
 
         let match: Match = try queue.sync { [unowned self] in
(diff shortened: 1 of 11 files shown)


Sources/DeepLinking/Router.swift
```diff
@@ -9,7 +9,9 @@
 import UIKit
 
 public protocol RouteHandler {
-    func handle(route: URL, parameters: [String : String], queryItems: [URLQueryItem])
+    associatedtype T
+
+    func handle(route: URL, parameters: [String : String], queryItems: [URLQueryItem], completion: ((T) -> Void)?)
```
jakewelton: Just a thought. Do you think it might be nice to assign a default value of `nil` to the `completion`? This way, the caller can completely miss that parameter if they don't want to use it?

jakewelton: Just a thought. Do you think it might be nice to assign a default value of `nil` to the `completion`? This way, the caller can completely miss that parameter if they don't want to use it?

EDIT: I guess the same question for the route function below too 🤔

Tests/AlicerceTests/DeepLinking/TreeRouterTests.swift
```diff
@@ -1115,7 +1116,7 @@ class TreeRouterTests: XCTestCase {
         let expectation = self.expectation(description: "TreeRouter.route")
         defer { waitForExpectations(timeout: expectationTimeout, handler: expectationHandler) }
 
-        testHandler.didCallHandle = { (url, parameters, queryItems) in
+        let completion: (HandledRoute) -> Void = { (url, parameters, queryItems) in
```
jakewelton: I don't think we need the brackets around the parameter list for the block? 😕 

Same comment for all the blocks below too 😄

Sources/DeepLinking/Router.swift
p4checo: Unfortunately we can't do that on protocols yet 😛 

However that's my intention for concrete implementations, yes 😉 

I now noticed that I didn't add that default parameter on `TestHandler`, but I think it was  because the `TreeRouter`'s `route` essentially already does that implicitly, since it's API has the `completion` parameter defaulting to `nil`. I can add it though 😄

jakewelton: Ah thats my bad. Sorry I didn't notice it was a protocol. I just saw it in the diff and assumed it was a class 😞

Tests/AlicerceTests/DeepLinking/TreeRouterTests.swift
p4checo: You're right. Don't know why I added them initially 😇 

Will remove 😉

Sources/DeepLinking/Router.swift
p4checo: Another important aspect is that the user isn't supposed to call `Handler.handle` directly. That's the `TreeRouter`'s responsibility, and as such having a default parameter on implementations isn't relevant since `TreeRouter.route` simply forwards whatever `completion` it receives to the respective handler. 😉

jakewelton: Ohh okay I see! That makes sense 👍

PR #77, opened 2017-08-16
Fix comparison of Page and Event

diogoeiras:
* Fixing comparison between different Pages and Events, where the 'left' name should be compared to the 'right' name

* Added some tests to Page and Event


p4checo:
Hi @diogoeiras, are you going to update this PR? 😉

PR #80, opened 2017-09-19
JSON date parse helper 📅 ⏰

portellaa:
Created helpers to parse and format a date from a JSON

Type on JSON is inferred from the formatter closure input type.

There is also an Optional version of the API.

### DIFF
--- a/Sources/Network/JSON.swift
+++ b/Sources/Network/JSON.swift
@@ -278,15 +278,18 @@ public enum JSON {
     ///   - parseAPIError: The API error parsing closure.
     /// - Returns: The Date obtained from the formatter.
     /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
-    public static func parseDateAttribute<T>(key: JSON.AttributeKey,
+    public static func parseDateAttribute<T>(_ key: JSON.AttributeKey,
                                              json: JSON.Dictionary,
                                              formatter: (T) -> Date?,
                                              parseAPIError: ParseAPIErrorClosure? = nil) throws -> Date {
 
-        guard let parsedDate = try parseOptionalDateAttribute(key: key,
-                                                          json: json,
-                                                          formatter: formatter,
-                                                          parseAPIError: parseAPIError) else {
+        guard let rawValue = json[key] else {
+            throw parseAPIError?(json) ?? Error.missingAttribute(key, json: json)
+        }
+
+        let jsonValue: T = try parseValue(rawValue: rawValue, key: key, json: json, parseAPIError: parseAPIError)
+
+        guard let parsedDate = formatter(jsonValue) else {
             throw Error.unexpectedAttributeValue(key, json: json)
         }
 
@@ -306,13 +309,14 @@ public enum JSON {
     ///   - parseAPIError: The API error parsing closure.
     /// - Returns: The Date obtained from the formatter or nil.
     /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
-    public static func parseOptionalDateAttribute<T>(key: JSON.AttributeKey,
+    public static func parseOptionalDateAttribute<T>(_ key: JSON.AttributeKey,
                                                      json: JSON.Dictionary,
                                                      formatter: (T) -> Date?,
                                                      parseAPIError: ParseAPIErrorClosure? = nil) throws -> Date? {
 
         guard let rawValue = json[key] else {
-            throw parseAPIError?(json) ?? Error.missingAttribute(key, json: json)
+            if let apiError = parseAPIError?(json) { throw apiError }
+            return nil
         }
 
         let jsonValue: T = try parseValue(rawValue: rawValue, key: key, json: json, parseAPIError: parseAPIError)
--- a/Sources/Network/JSON.swift
+++ b/Sources/Network/JSON.swift
@@ -321,7 +321,11 @@ public enum JSON {
 
         let jsonValue: T = try parseValue(rawValue: rawValue, key: key, json: json, parseAPIError: parseAPIError)
 
-        return formatter(jsonValue)
+        guard let date = formatter(jsonValue) else {
+            throw Error.unexpectedAttributeValue(key, json: json)
+        }
+
+        return date
     }
 
     // MARK: - Private methods
--- a/Sources/Network/JSON.swift
+++ b/Sources/Network/JSON.swift
@@ -264,6 +264,62 @@ public enum JSON {
         return try parseOptionalRawRepresentableAttribute(T.self, key: key, json: json, parseAPIError: parseAPIError)
     }
 
+
+    /// Parse an attribute of type `T` with the given key on the given JSON dictionary, trying to convert it to a date,
+    /// using the provided formatter which receives a `T` and returns a Date or nil if it fails.
+    /// An exception is thrown if the formatter is unable to convert the provided value into Date.
+    /// Additionally, an optional `parseAPIError` closure can be supplied so that if a parsing step fails
+    /// an attempt is made to extract a domain specific error and throw it.
+    ///
+    /// - Parameters:
+    ///   - key: The JSON attribute key to parse.
+    ///   - json: The JSON dictionary.
+    ///   - formatter: The formatter to convert the parsed type into Date.
+    ///   - parseAPIError: The API error parsing closure.
+    /// - Returns: The Date obtained from the formatter.
+    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
+    public static func parseDateAttribute<T>(key: JSON.AttributeKey,
+                                             json: JSON.Dictionary,
+                                             formatter: (T) -> Date?,
+                                             parseAPIError: ParseAPIErrorClosure? = nil) throws -> Date {
+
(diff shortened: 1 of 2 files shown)


Sources/Network/JSON.swift
```diff
@@ -262,6 +262,62 @@ public enum JSON {
         return try parseOptionalRawRepresentableAttribute(T.self, key: key, json: json, parseAPIError: parseAPIError)
     }
 
+
```
p4checo: you forgot the specified type variants 😇

Sources/Network/JSON.swift
```diff
@@ -262,6 +262,62 @@ public enum JSON {
...
+    ///   - parseAPIError: The API error parsing closure.
+    /// - Returns: The Date obtained from the formatter.
+    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
+    public static func parseDateAttribute<T>(key: JSON.AttributeKey,
+                                             json: JSON.Dictionary,
+                                             formatter: (T) -> Date?,
+                                             parseAPIError: ParseAPIErrorClosure? = nil) throws -> Date {
+
+        guard let parsedDate = try parseOptionalDateAttribute(key: key,
+                                                          json: json,
```
p4checo: bad alignment 🚓 💄

Sources/Network/JSON.swift
```diff
@@ -262,6 +262,62 @@ public enum JSON {
...
+    ///   - parseAPIError: The API error parsing closure.
+    /// - Returns: The Date obtained from the formatter or nil.
+    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
+    public static func parseOptionalDateAttribute<T>(key: JSON.AttributeKey,
+                                                     json: JSON.Dictionary,
+                                                     formatter: (T) -> Date?,
+                                                     parseAPIError: ParseAPIErrorClosure? = nil) throws -> Date? {
+
+        guard let rawValue = json[key] else {
+            throw parseAPIError?(json) ?? Error.missingAttribute(key, json: json)
```
p4checo: this should simply return `nil` since the attribute is optional, or throw the API Error if one is returned by the `parseAPIError` 

like: 
```swift
        guard let rawValue = json[key] else {
            if let apiError = parseAPIError?(json) { throw apiError }
            return nil
        }
```

Sources/Network/JSON.swift
```diff
@@ -262,6 +262,62 @@ public enum JSON {
...
+        }
+
+        let jsonValue: T = try parseValue(rawValue: rawValue, key: key, json: json, parseAPIError: parseAPIError)
+
+        return formatter(jsonValue)
```
p4checo: What if the formatter fails? Shouldn't we throw an exception? 

e.g.: `throw Error.unexpectedAttributeValue(key, json: json)`

Sources/Network/JSON.swift
```diff
@@ -262,6 +262,62 @@ public enum JSON {
...
+    /// an attempt is made to extract a domain specific error and throw it.
+    ///
+    /// - Parameters:
+    ///   - key: The JSON attribute key to parse.
+    ///   - json: The JSON dictionary.
+    ///   - formatter: The formatter to convert the parsed type into Date.
+    ///   - parseAPIError: The API error parsing closure.
+    /// - Returns: The Date obtained from the formatter.
+    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
+    public static func parseDateAttribute<T>(key: JSON.AttributeKey,
```
p4checo: on these API variants the `key` parameter is omitted `_ key:` 😉

Tests/AlicerceTests/Network/JSONTests.swift
```diff
@@ -1531,6 +1531,121 @@ class JSONTests: XCTestCase {
...
+        }(DateFormatter())
+
+        let json = [
+            "testDate" : dateFormatter.string(from: testDate)
+        ]
+
+        do {
+            let date = try JSON.parseDateAttribute(key: "testDate", json: json, formatter: dateFormatter.date)
+
+            XCTAssertEqual(testDate, date)
```
p4checo: we could assert `assertDumpsEqual` on the `json`s 😛

Tests/AlicerceTests/Network/JSONTests.swift
```diff
@@ -1531,6 +1531,121 @@ class JSONTests: XCTestCase {
...
+            let date = try JSON.parseOptionalDateAttribute(key: "testDate",
+                                                           json: json,
+                                                           formatter: Date.init(timeIntervalSince1970:))
+
+            XCTAssertNil(date)
+        } catch {
+            XCTFail("🔥: unexpected error \(error)")
+        }
+    }
+
```
p4checo: There's a missing test where the raw value is right, but the conversion fails with an error (which currently returns `nil`)

Sources/Network/JSON.swift
```diff
@@ -304,13 +307,14 @@ public enum JSON {
...
                                                      formatter: (T) -> Date?,
                                                      parseAPIError: ParseAPIErrorClosure? = nil) throws -> Date? {
 
         guard let rawValue = json[key] else {
-            throw parseAPIError?(json) ?? Error.missingAttribute(key, json: json)
+            if let apiError = parseAPIError?(json) { throw apiError }
+            return nil
         }
 
         let jsonValue: T = try parseValue(rawValue: rawValue, key: key, json: json, parseAPIError: parseAPIError)
```
p4checo: missing the above check here on the `formatter()` call too 😉

Sources/Network/JSON.swift
portellaa: But this is the case where it can be Optional, right? so the formatter may not be able to convert it, fair enough, but it returns nil. 
it doesn't seems wrong to me

Sources/Network/JSON.swift
portellaa: what do you mean? should we call the `apiError` when the formatter fails?

Tests/AlicerceTests/Network/JSONTests.swift
portellaa: this one 👉 `testParseDateAttribute_WithInvalidStringForFormatter_ShouldFailWithUnexpectedAttributeValue`

Sources/Network/JSON.swift
p4checo: From my perspective, the `parseOptional` should only return `nil` if the key doesn't come in the JSON at all, and not in any other case.

So, if the key is present in the JSON, we should treat it as a normal `parseAttribute`, and in this case if the `formatter()` call fails and returns `nil`, it means that the data was invalid and we should notify the caller. Makes sense? 😛

Tests/AlicerceTests/Network/JSONTests.swift
p4checo: Yes, I think so 😉
Also, we should do the same for the `parseOptionalDateAttribute` variant 😇

p4checo: Yes, it is 😉 
Still, we should do the same test for the `parseOptionalDateAttribute` variant 😇

Sources/Network/JSON.swift
p4checo: From my understanding, the `parseOptionalAttribute` should only return `nil` if the JSON doesn't contain a value for that key. If it does contain the value, it should essentially behave like a normal `parseAttribute`.

So in this case, I think we should preserve the behavior from the other one in regards to invoking `formatter()` 😜

portellaa: agree ✅

Tests/AlicerceTests/Network/JSONTests.swift
```diff
@@ -1531,6 +1531,155 @@ class JSONTests: XCTestCase {
...
+        let json = [
+            "testDate" : testDate.timeIntervalSince1970
+        ]
+
+        do {
+            let date = try JSON.parseDateAttribute("testDate",
+                                                   json: json,
+                                                   formatter: Date.init(timeIntervalSince1970:))
+
+            XCTAssertEqual(date.timeIntervalSince1970, testDate.timeIntervalSince1970)
```
p4checo: I know it should be equivalent, but shouldn't we compare `date` and `testDate` directly? 💅

Tests/AlicerceTests/Network/JSONTests.swift
```diff
@@ -1531,6 +1531,155 @@ class JSONTests: XCTestCase {
...
+            "testDate" : testDate.timeIntervalSince1970
+        ]
+
+        do {
+            let date = try JSON.parseOptionalDateAttribute("testDate",
+                                                           json: json,
+                                                           formatter: Date.init(timeIntervalSince1970:))
+
+            XCTAssertNotNil(date)
+            XCTAssertEqual(date?.timeIntervalSince1970, testDate.timeIntervalSince1970)
```
p4checo: same as above

Tests/AlicerceTests/Network/JSONTests.swift
portellaa: yeeaaahhhhh it makes sense 🤔 ok

portellaa:
I tried this, but it fails, even if the printed date is the same 😳 so i will leave this with the `timeIntervalSince1970` is that ok? @p4checo

p4checo:
Fine by me 👌😉
🚀

PR #81, opened 2017-09-22
Add helper methods and fix some NIB related code

pbaltarejo:
Fix `NibView` extension so that the `nib` property correctly return the `UINib` instance of the view with the same name

The protocol `ReusableView` don't need to implement the protocol `View`.
For example, if the view is implemented using Interface Builder all subviews and constraints are setted in IB. No need to implement `setUpSubviews()` and `setUpConstraints()`. However this view should be allowed to implement the `ReusableView` protocol.

Add helper register method in `UICollectionView` and `UITableView` to register cells that are implemented in Interface Builder

### DIFF
--- a/Sources/View/ReusableView.swift
+++ b/Sources/View/ReusableView.swift
@@ -25,9 +25,19 @@ public extension UICollectionView {
         register(cellType, forCellWithReuseIdentifier: cellType.reuseIdentifier)
     }
 
+    func register<T: UICollectionViewCell>(_ cellType: T.Type)
+    where T: ReusableView, T: NibView {
+        register(cellType.nib, forCellWithReuseIdentifier: cellType.reuseIdentifier)
+    }
+
     func register<T: UICollectionReusableView>(_ viewType: T.Type, forSupplementaryViewOfKind kind: String)
     where T: ReusableView {
         register(viewType, forSupplementaryViewOfKind: kind, withReuseIdentifier: viewType.reuseIdentifier)
+    }
+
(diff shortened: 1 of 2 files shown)


portellaa:
It seems that we are missing Unit tests for the UINib helpers 😫

Sources/View/ReusableView.swift
```diff
@@ -8,7 +8,7 @@
 
 import UIKit
 
-public protocol ReusableView: View {
```
p4checo: If we remove this conformance on `ReusableView`, I think we should add it to `ReusableViewModelView`

Sources/View/ReusableView.swift
```diff
@@ -25,11 +25,21 @@ public extension UICollectionView {
         register(cellType, forCellWithReuseIdentifier: cellType.reuseIdentifier)
     }
 
+    func register<T: UICollectionViewCell>(_ cellType: T.Type)
+    where T: ReusableView, T: NibView {
```
p4checo: We can probably use the new `&` operator on conformances: `T: ReusableView & NibView` 😉

portellaa:
@pbaltarejo do you intend to continue supporting this PR or can we close it?

I don't know what is the status on this, but there are some unfixed conflicts and comments to be done.

If not, we will the close the PR in the meanwhile, sorry 😓

p4checo:
Closing this since there hasn't been any activity on it for quite some time 😢 

@pbaltarejo, please feel free to re-open this if you need this in the future 😉

PR #82, opened 2017-09-28
Make the Resource capable of specifying if he wants the data to be retrieved first from the network or from the persistence

pbaltarejo:


Sources/Stores/Store.swift
```diff
@@ -42,8 +42,81 @@ private final class StoreCancelable: Cancelable {
...
+    @discardableResult
+    private func fetchNetworkFirst<Resource: NetworkResource & PersistableResource>(resource: Resource,
+                                                                                    _ completion: @escaping StoreCompletionClosure<T>)
+    -> Cancelable
+    where Resource.T == T {
+
+        let cancelable = StoreCancelable()
+
+        // Method to process the data when it's retrieved
+        func process(_ data: Data, fromCache: Bool) {
```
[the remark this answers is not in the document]
pbaltarejo: Yeah I agree, didn't want to refactor the other method that is proven being used, but I agree that would be better to refactor this class, will do it ;-)

Sources/Stores/Store.swift
```diff
@@ -42,8 +42,81 @@ private final class StoreCancelable: Cancelable {
...
+
+            do {
+                // update persistence with new value
+                self.persist(data, for: resource)
```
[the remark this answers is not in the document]
pbaltarejo: Good point, thanks ;)

Sources/Stores/Store.swift
```diff
@@ -42,8 +42,81 @@ private final class StoreCancelable: Cancelable {
...
+            if let error = error {
+
+                // 2nd - Fetch data from the Persistence
+                self.getPersistedData(for: resource) { (data) in
+
+                    // If we don't have on disk return the network error
+                    guard let data = data else { return completion(nil, error, false) }
+
+                    // Check if it's cancelled
+                    guard cancelable.isCancelled == false else { return completion(nil, .cancelled, false) }
```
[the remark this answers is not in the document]
pbaltarejo: yep, will change ;-)

Sources/Resource/PriorityResource.swift
```diff
@@ -0,0 +1,31 @@
...
+//  PriorityResource.swift
+//  Alicerce
+//
+//  Created by Pedro Baltarejo on 28/09/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public enum PriorityMode {
```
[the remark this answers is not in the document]
pbaltarejo: Thanks @p4checo. Really like your suggestion ;-)

Tests/AlicerceTests/Stores/StoreTestCase.swift
```diff
@@ -344,3 +350,161 @@ class StoreTestCase: XCTestCase {
         }
     }
 }
+
+class StoreTestCasePriority: XCTestCase {
```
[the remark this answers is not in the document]
pbaltarejo: Just trying to separate concepts. This tests we're to test if the order was being respected and not if the parser would work, for example. But will try to merge all of this tests

Tests/AlicerceTests/Stores/StoreTestCase.swift
```diff
@@ -39,6 +40,7 @@ class StoreTestCase: XCTestCase {
 
     private lazy var testResource: MockResource = {
         return MockResource(value: self.testValue,
+                            priority: .persistence,
```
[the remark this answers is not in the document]
pbaltarejo: Yeah, will try to add more tests

Tests/AlicerceTests/Stores/StoreTestCase.swift
```diff
@@ -344,3 +350,161 @@ class StoreTestCase: XCTestCase {
...
+            XCTAssertFalse(isCached)
+
+            defer { expectation.fulfill() }
+
+            guard let value = value else {
+                return XCTFail("🔥: missing value!")
+            }
+
+            XCTAssertEqual(value, self.testValueNetwork)
+            XCTAssertNotEqual(value, self.testValuePersistence)
```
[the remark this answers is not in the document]
pbaltarejo: Yeah, noticing that it's kind of dumb :)

Sources/Stores/Store.swift
```diff
@@ -119,77 +95,101 @@ public extension Store {
...
+    -> Cancelable {
+
+        return networkStack.fetch(resource: resource) { (inner: () throws -> Data) -> Void in
+            do {
+                let data = try inner()
+
+                completion(data, nil)
+
+            } catch let Network.Error.url(error as NSError) where error.domain == NSURLErrorDomain
+                && error.code == NSURLErrorCancelled {
```
p4checo: dat `&&` alignment 👀 🚓

Sources/Stores/Store.swift
```diff
@@ -119,77 +95,101 @@ public extension Store {
...
-                } catch let Network.Error.url(error as NSError) where error.domain == NSURLErrorDomain
-                                                                   && error.code == NSURLErrorCancelled {
-                    completion(nil, .cancelled, false)
-                } catch let error as Network.Error {
-                    completion(nil, .network(error), false)
-                } catch let error as Parse.Error {
-                    completion(nil, .parse(error), false)
-                } catch {
-                    completion(nil, .other(error), false)
+                    // If we got an error we need to check if we have the data on disk
```
p4checo: This comment is misleading

Sources/Stores/Store.swift
```diff
@@ -119,77 +95,101 @@ public extension Store {
...
-            do {
-                let value = try resource.parser(data)
-                completion(value, nil, true)
-            } catch {
-                // try to fetch fresh data if parsing of existent data failed
-                // TODO: remove from persistence?
-                return fetch(resource: resource)
+    // MARK: Processing Methods
+
+    private func process<Resource: NetworkResource & PersistableResource>(_ data: Data,
```
p4checo: this `process` is essentially a `parseAndPersist`, right? Not sure if you agree, but at least we would have something a bit more specific than just `process` 😆

Tests/AlicerceTests/Network/MockNetworkStack.swift
```diff
@@ -29,8 +29,7 @@ final class MockNetworkStack: NetworkStack {
     var beforeFetchCompletionClosure: (() -> Void)?
     var afterFetchCompletionClosure: (() -> Void)?
 
-    init(mockData: Data?, mockError: Network.Error?, queue: DispatchQueue = DispatchQueue.global()) {
-        precondition(mockData != nil || mockError != nil)
+    init(mockData: Data? = nil, mockError: Network.Error? = nil, queue: DispatchQueue = DispatchQueue.global()) {
```
p4checo: if both `mockData` and `mockError` are `nil`, this doesn't make much sense right? (that's why there was a `precondition` on the `init`)

(it will also crash when used 😆, see below)

Since they are both `var`'s, perhaps it's better to simply remove them from the `init`, since they are both initialized as `nil` anyway.

As mentioned above, it will crash when not configured correctly, so it's fine I guess.

Sources/Stores/Store.swift
```diff
@@ -119,77 +95,101 @@ public extension Store {
...
+                    // If we got an error we need to check if we have the data on disk
+                    if let error = error {
+
+                        return completion(nil, error, false)
+
+                    } else if let data = data {
+
+                        // parse the new value from the data
+                        self.process(data, fromCache: false, resource: resource, cancelable: cancelable, completion)
+                    }
```
p4checo: I know it's supposed to **never™** happen, but I would prefer if we added an `else` clause where we trigger a `fatalError`, since if that happens (due to any bug), the `completion` will never be called, for instance.

Sources/Stores/Store.swift
```diff
@@ -94,20 +76,14 @@ public extension Store {
...
 
             } else if let data = data {
 
-                // Check if it's cancelled
-                guard cancelable.isCancelled == false else { return completion(nil, .cancelled, false) }
-
                 // parse the new value from the data
-                process(data, fromCache: false)
+                self.process(data, fromCache: false, resource: resource, cancelable: cancelable, completion)
             }
```
p4checo: same as below (perhaps add a `fatalError`)

Tests/AlicerceTests/Stores/StoreTestCase.swift
```diff
@@ -76,18 +86,58 @@ class StoreTestCase: XCTestCase {
 
     // MARK: Failure
 
+    //     Network Stack: Error
+    // Persistence Stack: Error
```
p4checo: this is misleading, since the `noObjectForKey` is an error which symbolizes normal functioning of the persistence stack ( it's a simple cache miss 😛 )

same for some of the other cases below

Tests/AlicerceTests/Network/MockNetworkStack.swift
pbaltarejo: Agree. Will remove the `var`'s from `init`.

Sources/Stores/Store.swift
```diff
@@ -54,62 +54,156 @@ public struct StoreMetricsConfiguration<T> {
...
+    @discardableResult
+    private func fetchPersistenceFirst<Resource: NetworkResource & PersistableResource>(resource: Resource,
+                                                                                        _ completion: @escaping StoreCompletionClosure<T>)
+    -> Cancelable
+    where Resource.T == T {
+
+        let cancelable = StoreCancelable()
 
+        // 1st - Fetch data from the Persistence
+        self.getPersistedData(for: resource) { [weak self] (data) in
```
portellaa: 🚔 no need for `self.` here 💅 😝

Sources/Stores/Store.swift
```diff
@@ -54,62 +54,156 @@ public struct StoreMetricsConfiguration<T> {
...
+                                                                          cancelable: StoreCancelable,
+                                                                          _ completion: @escaping StoreCompletionClosure<T>)
+    where Resource.T == T {
+
+        do {
+            // Check if it's cancelled
+            guard cancelable.isCancelled == false else { return completion(nil, .cancelled, false) }
+
+            // parse the new value from the data
+            let value = try self.parse(data: data, for: resource)
```
portellaa: do we need `self.` here? 🤔

PR #87, opened 2017-11-14
Network handlers

portellaa:
Add ability to the network layer report activity to an upper layer


Sources/Network/RequestHandler.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Alicerce
+//
+//  Created by Luís Portela on 29/06/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol RequestHandler {
+    func handle(request: URLRequest)
```
p4checo: Not fully convinced about these new API names, sorry 👮 😇 

First, I see this more as a sort of `Interceptor` and not properly a `Handler`, since they aren't supposed to handle anything, right? I see them doing more logging/analytics related stuff, perhaps enriching the request with some metadata (e.g. associated objects), or something like that. Or am I wrong? 😛 

In that case, I see the API's as more of a will/did, e.g.

`willPerform(request: URLRequest)`
`didPerform(request: URLRequest, response: URLResponse?, error: Error?)`

or `willFetch/didFetch`

or a more explicit: 

`intercept(_ request: URLRequest)`
`interceptResponse(from request: URLRequest, response: URLResponse?, error: Error?)`

(or similar)

Also, would it make sense to also intercept the response `Data`?

portellaa: I understand, i also don't like the naming. 

This will not mutate/enrich the request or any received parameter, like you correctly said this is more logging/analytics.

I believe Interceptor seems an awesome name 😊 Are you comfortable with that?

Yes, we should send the data off course, i forgot as usual 😂

Sources/Network/RequestInterceptor.swift
```diff
@@ -0,0 +1,14 @@
...
+//  RequestInterceptor.swift
+//  Alicerce
+//
+//  Created by Luís Portela on 29/06/2017.
+//  Copyright © 2017 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol RequestInterceptor {
```
p4checo: 😍

PR #89, opened 2017-11-28
Changes the analytics related initialisers to be public

mschuetz72:
Just like the title says, the Analytics related classes couldn't be used as the initialisers were `internal`. This PR changes the access level to `public`

### DIFF
--- a/Sources/Analytics/Configuration.swift
+++ b/Sources/Analytics/Configuration.swift
@@ -19,7 +19,7 @@ public extension Analytics {
         let queueQoS: DispatchQoS
         let extraParameters: Parameters?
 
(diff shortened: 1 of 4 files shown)


Sources/Analytics/Configuration.swift
```diff
@@ -19,7 +19,7 @@ public extension Analytics {
         let queueQoS: DispatchQoS
         let extraParameters: Parameters?
 
-        init(queueQoS: DispatchQoS = .default,
+        public init(queueQoS: DispatchQoS = .default,
              extraParameters: Parameters? = nil) {
```
p4checo: bad alignment 🚓 😆

PR #90, opened 2017-11-30
Changes the analytics related attributes to be public

mschuetz72:
Public strikes again 🤕


p4checo:
It never gets old... 🤣

PR #91, opened 2017-12-15
Performance Metrics Layer ✨

portellaa:
First draft.

Yes, lot to do yet and no time for bit explanation 😛 sorry


portellaa:
🤔 measure block is nice, but it doesn't works very well with async code, unless the closure parameter is the identifier or we offer another closure to end the timer

mschuetz72:
If needed in future, there could be `startTimer(uniqueKey: String)`, `pauseTimer(uniqueKey)` and `stopTimer(uniqueKey)` methods to overcome the block based limitations. just a thought, so.

p4checo:
I think we can add public `begin/end` API's, if the user doesn't want to use closure based API's, but I don't see a use for `pauseTimer` 🤔

p4checo:
Just so that we don't forget in the future, we should eventually synchronize access to `trackers`.
It shouldn't be a major problem as-is because these things are typically configured on startup and never changed, but we should strive for correctness anyway 💪😄

PR #92, opened 2017-12-15
Measure my store 📏 🙄

portellaa:


Sources/Stores/Store.swift
```diff
@@ -85,8 +86,10 @@ public extension Store {
             }
 
             do {
-                let value = try resource.parser(data)
-                completion(value, nil, true)
+                try self.metrics?.measure(with: "Parse of \(T.self) with size: \(data.endIndex)") {
```
p4checo: we could make the key configurable by injecting a `parseMetricKey(typeName: String, dataSize: Int)`
or `parseMetricKey(type: Any.Type, dataSize: Int)` which would default to `"Parse of \(type) with size: \(dataSize)"`

Sources/Stores/Store.swift
```diff
@@ -85,8 +86,10 @@ public extension Store {
...
+                    let value = try resource.parser(data)
+                    completion(value, nil, true)
```
p4checo: this should be outside the closure 😉

Sources/Stores/Store.swift
p4checo: Additionally, if `metrics` is `nil` the parse will never be made 💥 🙈 😆

PR #98, opened 2018-01-03
Keyboard Observer - Enable/Disable tap multi touch

danieladias:
## Description
- Allows tap gesture on keyboard observer to be configurable in terms of allowing multi touch.

### DIFF
--- a/Sources/Observers/KeyboardObserver.swift
+++ b/Sources/Observers/KeyboardObserver.swift
@@ -13,13 +13,20 @@ public final class KeyboardObserver: NSObject {
     fileprivate var isKeyboardVisible = false
 
     private weak var window: UIWindow?
+    private let tapGestureRecognizer = UITapGestureRecognizer()
+
+    var shouldTapCancelTouches: Bool = true {
+        didSet {
+            tapGestureRecognizer.cancelsTouchesInView = shouldTapCancelTouches
+        }
+    }
 
     public init(window: UIWindow) {
         self.window = window
 
         super.init()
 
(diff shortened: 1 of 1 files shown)


Sources/Observers/KeyboardObserver.swift
```diff
@@ -13,13 +13,20 @@ public final class KeyboardObserver: NSObject {
     fileprivate var isKeyboardVisible = false
 
     private weak var window: UIWindow?
+    private let tapGestureRecognizer = UITapGestureRecognizer()
```
p4checo: Since we are now retaining the `UITapGestureRecognizer`, I think we will be implicitly retaining the `window` as well, right? (since we add it to the window and the window retains it)

So, if we want to preserve the weak reference to the window, perhaps we should also store the `UITapGestureRecognizer` as a weak reference. Alternatively, make `window` a strong reference so that the actual situation is more explicit.

What to you guys think? 🤔

portellaa: 🤔 perhaps not a bad idea. because previously we were not storing the gesture recognizer, only the window retained it.

I agree, nice catch ✅ 👍

danieladias: Yes, that is true. To be honest I didn't want to store the property. For me ideally the `cancelTouchesInView` should always false but if we want to make it configurable it has to be stored.

p4checo: On another note, I think we should remove the `UITapGestureRecognizer` from the `window` on `deinit` since we are only using it for this purpose ♻️

PR #101, opened 2018-01-16
Removed the writtenItems property from LogDestination protocol

mschuetz72:
The `writtenItems` property was exclusively used for unit test validations and did not make sense. Removed the property from protocol, from the `LogDestination`s as well as from the unit tests.


Tests/AlicerceTests/Logging/ColoredLevelTests.swift
```diff
@@ -90,16 +91,19 @@ class ColoredLevelTests: XCTestCase {
...
 
     //MARK:- private methods
 
-    private func logfileContent(logfileURL: URL) -> String {
+    private func logfileContent(logfileURL: URL) -> (String, Int) {
 
-        return (try? String(contentsOf: logfileURL)) ?? ""
+        let content = (try? String(contentsOf: logfileURL)) ?? ""
+        let numberOfLines = content.split(separator: "\n").count
+        return (content, numberOfLines)
```
p4checo: since the expected content already contains the `\n`'s, it should be enough to test the output, right?

Tests/AlicerceTests/Logging/ConsoleLogDestinationTests.swift
```diff
@@ -38,7 +40,12 @@ class ConsoleLogDestinationsTests: XCTestCase {
 
         // preparation of the test subject
 
-        let destination = Log.ConsoleLogDestination(minLevel: .verbose, queue: queue)
+        let outputClosure: Log.ConsoleLogDestination.OutputClosure = {
+            [weak self] _ in self?.writtenLines += 1 }
```
p4checo: this could be done with an `expectation`, and configuring its `expectedFulfillmentCount` 😉

PR #102, opened 2018-01-25
Improvement/bool extension

dbeleza:
This new extension will provide a `toggle()` method that will make code cleaner.

before
`instance.object1.enable = !instance.object1.enable`

after
`instance.object1.enable.toggle()`


portellaa:
@dbeleza can you add some tests please? they are quite simple :)

dbeleza:
Done @portellaa. 👍

p4checo:
@dbeleza can you please rebase so that you get the latest fixes on the JSON date tests, and the build passes? Thanks! 🙏

portellaa:
@dbeleza no need to rebase now, this will be fixed as soon as we merge the #60 branch.

portellaa:
@dbeleza go ahead rebase and merge this 😉 🍻

jakewelton:
Nice addition @dbeleza. Not sure if everyone is already aware, but the Swift language will actually implement this as part of Swift 4.2.

Proposal: https://github.com/apple/swift-evolution/blob/master/proposals/0199-bool-toggle.md
PR Implementing This: https://github.com/apple/swift/pull/14586

Just wanted to mention it in case it changes anything :smile:

p4checo:
Nice catch @jakewelton 👍

Since 4.2 is right around the corner (hopefully), perhaps we can close this PR since this functionality will essentially be duplicated real soon.

What do you guys think?

portellaa:
@dbeleza can you close this? This is now part of Swift 4.2 so no need to have this on Alicerce, at least not like this.

we can discuss if it's needed to support this for old swift versions, but i think we don't want to follow that path.

Just opening the discussion to it 😄

PR #103, opened 2018-02-15
Helper to always set `translatesAutoresizingMaskIntoConstraints` to false

beloso:
This is a helper method that will ensure that `translatesAutoresizingMaskIntoConstraints` is always set to `false` when adding layout constraints manually.

### DIFF
--- a/Sources/UIKit/NSLayoutConstraint.swift
+++ b/Sources/UIKit/NSLayoutConstraint.swift
@@ -11,6 +11,11 @@ import UIKit
 
 extension NSLayoutConstraint {
 
+
+    /// Similar to NSLayoutConstraint.activate(), but this method ensures that the views
+    /// that are adding the constraints disable translatesAutoresizingMaskIntoConstraints
+    ///
+    /// - Parameter constraints: an array of constrains to activate
     public static func add(constraints: [NSLayoutConstraint]) {
 
         constraints.forEach {
--- a/Sources/UIKit/NSLayoutConstraint.swift
+++ b/Sources/UIKit/NSLayoutConstraint.swift
@@ -9,22 +9,40 @@
 import Foundation
 import UIKit
 
-extension NSLayoutConstraint {
-
+extension UIView {
 
     /// Similar to NSLayoutConstraint.activate(), but this method ensures that the views
     /// that are adding the constraints disable translatesAutoresizingMaskIntoConstraints
     ///
     /// - Parameter constraints: an array of constrains to activate
-    public static func add(constraints: [NSLayoutConstraint]) {
+    public static func activateConstraints(_ constraints: [NSLayoutConstraint]) {
 
         constraints.forEach {
-            if let v = $0.firstItem as? UIView {
 
-                v.translatesAutoresizingMaskIntoConstraints = false
+            if let view = $0.firstItem as? UIView {
+
+                view.translatesAutoresizingMaskIntoConstraints = false
             }
         }
 
-        activate(constraints)
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    func edgesToView(_ view: UIView, insets: UIEdgeInsets = .zero) {
+
+        UIView.activateConstraints([
+            topAnchor.constraint(equalTo: view.topAnchor, constant: insets.top),
+            leftAnchor.constraint(equalTo: view.leftAnchor, constant: insets.left),
+            bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: -insets.bottom),
+            rightAnchor.constraint(equalTo: view.rightAnchor, constant: -insets.right)
+            ])
+    }
+
+    func centerInView(_ view: UIView, offset: UIOffset = .zero) {
+
+        UIView.activateConstraints([
+            centerXAnchor.constraint(equalTo: view.centerXAnchor, constant: offset.horizontal),
+            centerYAnchor.constraint(equalTo: view.centerYAnchor, constant: offset.vertical)
+            ])
     }
 }
--- a/Sources/UIKit/NSLayoutConstraint.swift
+++ b/Sources/UIKit/NSLayoutConstraint.swift
@@ -1,0 +1,25 @@
+//
+//  NSConstraintLayout.swift
+//  Alicerce
+//
+//  Created by Tiago Veloso on 15/02/2018.
+//  Copyright © 2018 Mindera. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+extension NSLayoutConstraint {
+
+    public static func add(constraints: [NSLayoutConstraint]) {
+
+        constraints.forEach {
+            if let v = $0.firstItem as? UIView {
+
+                v.translatesAutoresizingMaskIntoConstraints = false
+            }
+        }
+
+        activate(constraints)
+    }
+}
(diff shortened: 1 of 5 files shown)


beloso:
Added a unit test

Sources/UIKit/NSLayoutConstraint.swift
```diff
@@ -0,0 +1,30 @@
...
+import UIKit
+
+extension NSLayoutConstraint {
+
+
+    /// Similar to NSLayoutConstraint.activate(), but this method ensures that the views
+    /// that are adding the constraints disable translatesAutoresizingMaskIntoConstraints
+    ///
+    /// - Parameter constraints: an array of constrains to activate
+    public static func add(constraints: [NSLayoutConstraint]) {
```
p4checo: I think `add` is not clear enough as to what this API does, since we're essentially activating the constraints and introducing a side effect (marking the `translatesAutoresizingMaskIntoConstraints` flag as `false`).

I know this is more verbose, but perhaps:
`func activateAndDisableAutoresizingMasksTranslation(_ constraints: [NSLayoutConstraint])`?

Or, we could make a twist and make disabling the masks a parameter:
`func activate(_ constraints: [NSLayoutConstraint], disableAutoresizingMasksTranslation: Bool = true)`

What do you say?

Sources/UIKit/NSLayoutConstraint.swift
```diff
@@ -0,0 +1,30 @@
...
+
+        constraints.forEach {
+            if let v = $0.firstItem as? UIView {
+
```
p4checo: please be consistent with vertical whitespaces 💄

Sources/UIKit/NSLayoutConstraint.swift
portellaa: `activate` is the instance method, which for me, nhec 😕

Sources/UIKit/NSLayoutConstraint.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Alicerce
+//
+//  Created by Tiago Veloso on 15/02/2018.
+//  Copyright © 2018 Mindera. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+extension NSLayoutConstraint {
```
portellaa: wouldn't it be better if we add the method to `UIView`? 🤔

Sources/UIKit/NSLayoutConstraint.swift
portellaa: 💄  can we call it `let view` instead of `let v` ?

beloso:
I've made some changes to account for your suggestions. I've also added two helper methods, but I don't know how to test them. Do you guys have any idea about hot to test them?

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,48 @@
...
+            if let view = $0.firstItem as? UIView {
+
+                view.translatesAutoresizingMaskIntoConstraints = false
+            }
+        }
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    func edgesToView(_ view: UIView, insets: UIEdgeInsets = .zero) {
```
p4checo: perhaps a better name could be `constrainEdgesToView(_ view: UIView, insets: UIEdgeInsets = .zero)`?

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,48 @@
...
+
+        UIView.activateConstraints([
+            topAnchor.constraint(equalTo: view.topAnchor, constant: insets.top),
+            leftAnchor.constraint(equalTo: view.leftAnchor, constant: insets.left),
+            bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: -insets.bottom),
+            rightAnchor.constraint(equalTo: view.rightAnchor, constant: -insets.right)
+            ])
+    }
+
+    func centerInView(_ view: UIView, offset: UIOffset = .zero) {
```
p4checo: like above, perhaps `constrainCenteredInView(_ view: UIView, offset: UIOffset = .zero)`?

nunogoncalves: Sorry to come out of nowhere but: 

`center(in:, with:)` reads pretty nicely. Since the types are self explanatory, unlike String or Int, then I think it makes a lot of sense. :)

`label.center(in: parentView, with: UIOffset(...)`

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Similar to NSLayoutConstraint.activate(), but this method ensures that the views
+    /// that are adding the constraints disable translatesAutoresizingMaskIntoConstraints
+    ///
+    /// - Parameter constraints: an array of constrains to activate
+    public static func activateConstraints(_ constraints: [NSLayoutConstraint]) {
+
+        constraints.forEach {
+
+            if let view = $0.firstItem as? UIView {
```
nunogoncalves: You could simply do `($0.firstItem as? UIView)?.translates....`

beloso:
Made some fixes, I need to test the visual effects of `left`, `right`, `top`, `bottom`.

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,79 @@
...
+        constraints.forEach {
+
+            ($0.firstItem as? UIView)?.translatesAutoresizingMaskIntoConstraints = false
+        }
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    @available(iOS 9.0, *)
+    public func top(ofView view: UIView, offset: CGFloat = 0.0) {
```
portellaa: what about `public func top(of view: UIView, offset: ...)` ? 💄 💅

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,79 @@
...
+
+        UIView.activateConstraints([
+            bottomAnchor.constraint(equalTo: view.topAnchor, constant: -offset)
+            ])
+    }
+
+    @available(iOS 9.0, *)
+    public func left(ofView view: UIView, offset: CGFloat = 0.0) {
```
portellaa: same as above

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,79 @@
...
+
+        UIView.activateConstraints([
+            rightAnchor.constraint(equalTo: view.leftAnchor, constant: -offset)
+            ])
+    }
+
+    @available(iOS 9.0, *)
+    public func bottom(ofView view: UIView, offset: CGFloat = 0.0) {
```
portellaa: same as above

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,79 @@
...
+
+        UIView.activateConstraints([
+            topAnchor.constraint(equalTo: view.bottomAnchor, constant: offset)
+            ])
+    }
+
+    @available(iOS 9.0, *)
+    public func right(ofView view: UIView, offset: CGFloat = 0.0) {
```
portellaa: same as above

Sources/UIKit/UIView.swift
```diff
@@ -0,0 +1,79 @@
...
+
+        UIView.activateConstraints([
+            leftAnchor.constraint(equalTo: view.rightAnchor, constant: offset)
+            ])
+    }
+
+    @available(iOS 9.0, *)
+    public func edges(toView view: UIView, insets: UIEdgeInsets = .zero) {
```
portellaa: same as above

PR #106, opened 2018-02-20
Added persistence performance metrics

danieladias:
## Description
- Added metadata dictionary to performance events
- Added performance metrics to read/write operations of the persistence stack

### DIFF
--- a/Sources/PerformanceMetrics/PerformanceMetrics.swift
+++ b/Sources/PerformanceMetrics/PerformanceMetrics.swift
@@ -28,10 +28,10 @@ public final class PerformanceMetrics {
     // Measurement API
 
     public func measure<T>(with identifier: Identifier,
-                           metadata: Metadata = [:],
+                           metadata: Metadata? = nil,
                            measureBlock: () throws -> T) rethrows -> T {
 
-        begin(with: identifier, metadata: metadata)
+        begin(with: identifier)
 
         let measureResult = try measureBlock()
 
@@ -41,23 +41,23 @@ public final class PerformanceMetrics {
     }
 
     public func measureAsync<T>(with identifier: Identifier,
-                                metadata: Metadata = [:],
+                                metadata: Metadata? = nil,
                                 measureBlock: (_ end: () -> Void) throws -> T) rethrows -> T {
         
         let end: () -> Void = { [weak self] in
             self?.end(with: identifier, metadata: metadata)
         }
 
-        begin(with: identifier, metadata: metadata)
+        begin(with: identifier)
 
         return try measureBlock(end)
     }
 
-    public func begin(with identifier: PerformanceMetrics.Identifier, metadata: Metadata = [:]) {
-        trackers.forEach { $0.begin(with: identifier, metadata: metadata) }
+    public func begin(with identifier: Identifier) {
+        trackers.forEach { $0.begin(with: identifier) }
     }
     
-    public func end(with identifier: PerformanceMetrics.Identifier, metadata: Metadata = [:]) {
+    public func end(with identifier: Identifier, metadata: Metadata? = nil) {
         trackers.forEach { $0.end(with: identifier, metadata: metadata) }
     }
 }
--- a/Sources/PerformanceMetrics/PerformanceMetrics.swift
+++ b/Sources/PerformanceMetrics/PerformanceMetrics.swift
@@ -11,6 +11,7 @@ import Foundation
 public final class PerformanceMetrics {
 
     public typealias Identifier = String
+    public typealias Metadata = [String : Any]
 
     private lazy var trackers = [PerformanceMetricsTracker]()
 
@@ -26,32 +27,37 @@ public final class PerformanceMetrics {
 
     // Measurement API
 
-    public func measure<T>(with identifier: Identifier, measureBlock: () throws -> T) rethrows -> T {
-        begin(with: identifier)
+    public func measure<T>(with identifier: Identifier,
+                           metadata: Metadata = [:],
+                           measureBlock: () throws -> T) rethrows -> T {
+
+        begin(with: identifier, metadata: metadata)
 
         let measureResult = try measureBlock()
 
-        end(with: identifier)
+        end(with: identifier, metadata: metadata)
 
         return measureResult
     }
 
     public func measureAsync<T>(with identifier: Identifier,
+                                metadata: Metadata = [:],
                                 measureBlock: (_ end: () -> Void) throws -> T) rethrows -> T {
+        
         let end: () -> Void = { [weak self] in
-            self?.end(with: identifier)
+            self?.end(with: identifier, metadata: metadata)
         }
 
-        begin(with: identifier)
+        begin(with: identifier, metadata: metadata)
 
         return try measureBlock(end)
     }
 
-    public func begin(with identifier: PerformanceMetrics.Identifier) {
-        trackers.forEach { $0.begin(with: identifier) }
+    public func begin(with identifier: PerformanceMetrics.Identifier, metadata: Metadata = [:]) {
+        trackers.forEach { $0.begin(with: identifier, metadata: metadata) }
     }
     
-    public func end(with identifier: PerformanceMetrics.Identifier) {
-        trackers.forEach { $0.end(with: identifier) }
+    public func end(with identifier: PerformanceMetrics.Identifier, metadata: Metadata = [:]) {
+        trackers.forEach { $0.end(with: identifier, metadata: metadata) }
     }
 }
--- a/Sources/PerformanceMetrics/PerformanceMetrics.swift
+++ b/Sources/PerformanceMetrics/PerformanceMetrics.swift
@@ -10,9 +10,6 @@ import Foundation
 
 public final class PerformanceMetrics {
 
-    public typealias Identifier = String
-    public typealias Metadata = [String : Any]
-
     private lazy var trackers = [PerformanceMetricsTracker]()
 
     public init() {}
@@ -27,8 +24,8 @@ public final class PerformanceMetrics {
 
     // Measurement API
 
-    public func measure<T>(with identifier: Identifier,
-                           metadata: Metadata? = nil,
+    public func measure<T>(with identifier: PerformanceMetricsTracker.Identifier,
+                           metadata: PerformanceMetricsTracker.Metadata? = nil,
                            measureBlock: () throws -> T) rethrows -> T {
 
         begin(with: identifier)
@@ -40,8 +37,8 @@ public final class PerformanceMetrics {
         return measureResult
     }
 
-    public func measureAsync<T>(with identifier: Identifier,
-                                metadata: Metadata? = nil,
+    public func measureAsync<T>(with identifier: PerformanceMetricsTracker.Identifier,
+                                metadata: PerformanceMetricsTracker.Metadata? = nil,
                                 measureBlock: (_ end: () -> Void) throws -> T) rethrows -> T {
         
         let end: () -> Void = { [weak self] in
@@ -53,11 +50,11 @@ public final class PerformanceMetrics {
         return try measureBlock(end)
     }
 
-    public func begin(with identifier: Identifier) {
+    public func begin(with identifier: PerformanceMetricsTracker.Identifier) {
         trackers.forEach { $0.begin(with: identifier) }
     }
     
-    public func end(with identifier: Identifier, metadata: Metadata? = nil) {
+    public func end(with identifier: PerformanceMetricsTracker.Identifier, metadata: PerformanceMetricsTracker.Metadata? = nil) {
         trackers.forEach { $0.end(with: identifier, metadata: metadata) }
     }
 }
--- a/Sources/PerformanceMetrics/PerformanceMetricsTracker.swift
+++ b/Sources/PerformanceMetrics/PerformanceMetricsTracker.swift
@@ -9,6 +9,6 @@
 import Foundation
 
 public protocol PerformanceMetricsTracker: class {
-    func begin(with identifier: PerformanceMetrics.Identifier)
-    func end(with identifier: PerformanceMetrics.Identifier)
+    func begin(with identifier: PerformanceMetrics.Identifier, metadata: PerformanceMetrics.Metadata)
+    func end(with identifier: PerformanceMetrics.Identifier, metadata: PerformanceMetrics.Metadata)
 }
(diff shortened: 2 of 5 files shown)


Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -127,11 +149,27 @@ public final class DiskMemoryPersistenceStack: PersistenceStack {
     // MARK: - NSCache Related Operations
 
     private func cachedData(for key: Persistence.Key) -> Data? {
-        return cache.object(forKey: key.nsString) as Data?
+
+        let identifier = "\(Constants.PerformanceMetrics.readMemoryKey) \(key)"
```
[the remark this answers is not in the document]
p4checo: Also thinking that we could perhaps simplify this by implementing these metrics on the `object(for:completion:)` method, and in each exit path we already know where the value came from, which can also be an important metric, no?

So, every time something hits our persistence, we will tell on a single metric:
1. how long it took
2. where was it returned from (.memory, .disk)
3. sizes of both caches

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -321,6 +380,16 @@ public final class DiskMemoryPersistenceStack: PersistenceStack {
     }
 }
 
+extension DiskMemoryPersistenceStack: NSCacheDelegate {
+
+    public func cache(_ cache: NSCache<AnyObject, AnyObject>, willEvictObject obj: Any) {
+
+        guard let data = obj as? Data else { return }
```
p4checo: why do we need this check? Additionally, we will only decrement the counter when it's `Data`

perhaps it could be an assertion (given we only deal with `Data` objects)?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -179,7 +229,16 @@ public final class DiskMemoryPersistenceStack: PersistenceStack {
...
-            completion { () }
+            completion { [weak self] in
+
+                guard let strongSelf = self else { return () }
+                
+                strongSelf.performanceMetrics?.end(with: identifier,
+                                                   metadata:
+                    [Constants.PerformanceMetrics.totalDiskKey : strongSelf.usedDiskSize])
+
+                return ()
```
p4checo: I think we don't need the `()` 😛

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -74,11 +88,19 @@ public final class DiskMemoryPersistenceStack: PersistenceStack {
     private let fileManager: FileManager = FileManager.default
 
     private var usedDiskSize: UInt64 = 0 // Size in bytes
+    private var usedMemorySize: UInt64 = 0 // Size in bytes
 
-    public init(configuration: Configuration) {
+    private let performanceMetrics: PerformanceMetrics?
+
+    public init(configuration: Configuration,
+                performanceMetrics: PerformanceMetrics? = nil) {
```
p4checo: I would perhaps suggest injecting a configuration with the keys to use, to avoid these being essentially hardcoded to the constants. 
It also enables configuring other stuff in the future, maybe.

Sources/PerformanceMetrics/PerformanceMetricsTracker.swift
```diff
@@ -9,6 +9,6 @@
 import Foundation
 
 public protocol PerformanceMetricsTracker: class {
-    func begin(with identifier: PerformanceMetrics.Identifier)
-    func end(with identifier: PerformanceMetrics.Identifier)
+    func begin(with identifier: PerformanceMetrics.Identifier, metadata: PerformanceMetrics.Metadata)
```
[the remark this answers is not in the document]
danieladias: Yes, definitely. Removed when we were discussing and forgot to add back.

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: I discussed this with @portellaa and we thought this could be an improvement done later when we have more information on different performance trackers. Is it worth doing now then?

Sources/PerformanceMetrics/PerformanceMetrics.swift
```diff
@@ -26,32 +27,37 @@ public final class PerformanceMetrics {
 
     // Measurement API
 
-    public func measure<T>(with identifier: Identifier, measureBlock: () throws -> T) rethrows -> T {
-        begin(with: identifier)
+    public func measure<T>(with identifier: Identifier,
+                           metadata: Metadata = [:],
```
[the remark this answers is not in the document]
danieladias: Done

Sources/PerformanceMetrics/PerformanceMetricsTracker.swift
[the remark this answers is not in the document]
danieladias: I thought we discussed that it would make sense for more of event driven performance trackers.

Sources/Persistence/DiskMemoryPersistenceStack.swift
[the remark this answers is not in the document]
danieladias: Again, something I asked @portellaa and we thought they wanted to know what exact operation was being done since they want to measure every insert.

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Because the delegate method sends Any.

Sources/PerformanceMetrics/PerformanceMetrics.swift
```diff
@@ -26,32 +27,37 @@ public final class PerformanceMetrics {
...
         let measureResult = try measureBlock()
 
-        end(with: identifier)
+        end(with: identifier, metadata: metadata)
 
         return measureResult
     }
 
     public func measureAsync<T>(with identifier: Identifier,
+                                metadata: Metadata = [:],
```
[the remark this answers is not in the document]
danieladias: Done

Sources/PerformanceMetrics/PerformanceMetrics.swift
```diff
@@ -26,32 +27,37 @@ public final class PerformanceMetrics {
...
 
-        begin(with: identifier)
+        begin(with: identifier, metadata: metadata)
 
         return try measureBlock(end)
     }
 
-    public func begin(with identifier: PerformanceMetrics.Identifier) {
-        trackers.forEach { $0.begin(with: identifier) }
+    public func begin(with identifier: PerformanceMetrics.Identifier, metadata: Metadata = [:]) {
```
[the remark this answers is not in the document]
danieladias: Done

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Fixed

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Fixed I think. Not sure it is what we had in mind but will commit soon.

Sources/Persistence/DiskMemoryPersistenceStack.swift
p4checo: 🤦‍♂️ 

perhaps we could still fire an assertion if we don't receive a `Data`, no?

Sources/PerformanceMetrics/PerformanceMetricsTracker.swift
```diff
@@ -9,6 +9,6 @@
 import Foundation
 
 public protocol PerformanceMetricsTracker: class {
-    func begin(with identifier: PerformanceMetrics.Identifier, metadata: PerformanceMetrics.Metadata)
-    func end(with identifier: PerformanceMetrics.Identifier, metadata: PerformanceMetrics.Metadata)
+    func begin(with identifier: PerformanceMetrics.Identifier)
+    func end(with identifier: PerformanceMetrics.Identifier, metadata: PerformanceMetrics.Metadata?)
```
p4checo: This just hit me now (and it was already like this), but we're tying this protocol to a concrete implementation (`PerformanceMetrics`) 😆 

My suggestion is that we either define those `typealias` inside the protocol, or use the original types 😉

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -14,17 +14,31 @@ public protocol DiskMemoryPersistenceStackDelegate: class {
...
+        let metrics: PerformanceMetrics
+        let memoryAttributeKey: String
+        let diskAttributeKey: String
+        let readMemoryEventKey: String
+        let writeMemoryEventKey: String
+        let readDiskEventKey: String
+        let writeDiskEventKey: String
+
+        public init(metrics: PerformanceMetrics,
+                    memoryAttributeKey: String,
```
p4checo: we could use the previous constant values as default values 💅

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -150,26 +164,24 @@ public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
...
+        persistencePerformanceMetrics?.metrics.end(with: identifier)
 
         return data
     }
 
     private func setCachedData(_ data: Data, for key: Persistence.Key) {
 
-        let identifier = "\(Constants.PerformanceMetrics.writeMemoryKey) \(key)"
+        let identifier = persistencePerformanceMetrics?.writeMemoryEventKey ?? ""
+        let attribute = persistencePerformanceMetrics?.memoryAttributeKey ?? ""
```
p4checo: to avoid this dance we could perhaps create helper methods on `persistencePerformanceMetrics` for each metric 💄

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -150,26 +164,24 @@ public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
...
     private func cachedData(for key: Persistence.Key) -> Data? {
 
-        let identifier = "\(Constants.PerformanceMetrics.readMemoryKey) \(key)"
+        let identifier = persistencePerformanceMetrics?.readMemoryEventKey ?? ""
 
-        performanceMetrics?.begin(with: identifier)
+        persistencePerformanceMetrics?.metrics.begin(with: identifier)
         let data = cache.object(forKey: key.nsString) as Data?
-        performanceMetrics?.end(with: identifier)
+        persistencePerformanceMetrics?.metrics.end(with: identifier)
```
p4checo: shouldn't/couldn't we also associate the blob's size?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -231,13 +243,15 @@ public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
...
-                guard let strongSelf = self else { return () }
+                guard let strongSelf = self,
+                    let persistencePerformanceMetrics = strongSelf.persistencePerformanceMetrics else { return }
                 
-                strongSelf.performanceMetrics?.end(with: identifier,
-                                                   metadata:
-                    [Constants.PerformanceMetrics.totalDiskKey : strongSelf.usedDiskSize])
+                persistencePerformanceMetrics.metrics.end(
+                    with: identifier,
+                    metadata: [persistencePerformanceMetrics.diskAttributeKey : strongSelf.usedDiskSize]
```
p4checo: shouldn't/couldn't we also associate the blob's size?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -193,7 +205,7 @@ public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
             }
 
             completion { [weak self] in
-                self?.performanceMetrics?.end(with: identifier)
+                self?.persistencePerformanceMetrics?.metrics.end(with: identifier)
```
p4checo: shouldn't/couldn't we also associate the blob's size?

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Fixed

Sources/PerformanceMetrics/PerformanceMetricsTracker.swift
danieladias: Fixed

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Fixed

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Fixed

Sources/Persistence/DiskMemoryPersistenceStack.swift
danieladias: Fixed

Sources/Persistence/PersistencePerformanceMetrics.swift
```diff
@@ -0,0 +1,78 @@
...
+        metrics.end(with: readMemoryEventKey)
+    }
+
+    public func beginWriteMemory() {
+        metrics.begin(with: writeMemoryEventKey)
+    }
+
+    public func endWriteMemory(blobSize: UInt64, memorySize: UInt64) {
+        metrics.end(with: writeMemoryEventKey, metadata: [blobSizeAttributeKey : blobSize,
+                                                          memoryAttributeKey: memorySize])
```
p4checo: inconsistent white space in dictionary 🚓

Sources/Persistence/PersistencePerformanceMetrics.swift
```diff
@@ -0,0 +1,78 @@
...
+        metrics.end(with: readDiskEventKey)
+    }
+
+    public func beginWriteDisk() {
+        metrics.begin(with: writeDiskEventKey)
+    }
+
+    public func endWriteDisk(blobSize: UInt64, diskSize: UInt64) {
+        metrics.end(with: writeDiskEventKey, metadata: [blobSizeAttributeKey : blobSize,
+                                                        diskAttributeKey: diskSize])
```
p4checo: same as above 👮

danieladias: damn
Fixed

Sources/Persistence/PersistencePerformanceMetrics.swift
danieladias: Fixed

PR #110, opened 2018-03-22
HTTP Response 204 No Content

danieladias:
## Description
- Added handling for http response 204 with no content by checking success 204 and `Resource.Local` expected void.
- Added empty closure on `Resource` to return in case of 204.

### DIFF
--- a/Sources/Network/Parse.swift
+++ b/Sources/Network/Parse.swift
@@ -61,6 +61,7 @@ public enum Parse {
     ///
     /// - Parameter data: Raw data
     /// - Throws: A Parse.Error that can be of type
+    ///   - `unexpectedData` if data is not empty as expected
     public static func void(data: Data) throws {
         guard data.isEmpty else {
             throw Error.unexpectedData
--- a/Sources/Network/Parse.swift
+++ b/Sources/Network/Parse.swift
@@ -13,6 +13,7 @@ public enum Parse {
     public enum Error: Swift.Error {
         case json(Swift.Error)
         case image
+        case noData
     }
 
     /// Converts raw data into a `T` object
@@ -54,5 +55,15 @@ public enum Parse {
         }
 
         return image
+    }
+
+    /// Parses empty data
+    ///
+    /// - Parameter data: Raw data
+    /// - Throws: A Parse.Error that can be of type
+    public static func void(data: Data) throws {
+        guard data.isEmpty else {
+            throw Error.noData
+        }
     }
 }
--- a/Sources/Network/Parse.swift
+++ b/Sources/Network/Parse.swift
@@ -13,7 +13,7 @@ public enum Parse {
     public enum Error: Swift.Error {
         case json(Swift.Error)
         case image
-        case noData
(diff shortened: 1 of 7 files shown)


Sources/Resource/NetworkResource.swift
```diff
@@ -11,6 +11,7 @@ import Foundation
 public protocol NetworkResource: Resource {
 
     var request: URLRequest { get }
+    var empty: Remote { get }
```
p4checo: it might make sense for this to be a `static var`, no?

Sources/Network/Parse.swift
```diff
@@ -55,4 +56,14 @@ public enum Parse {
 
         return image
     }
+
+    /// Parses empty data
+    ///
+    /// - Parameter data: Raw data
+    /// - Throws: A Parse.Error that can be of type
```
p4checo: incomplete comment 😆 (if we keep the error)

Sources/Network/Parse.swift
```diff
@@ -55,4 +56,14 @@ public enum Parse {
...
+    public static func void(data: Data) throws {
```
p4checo: As discussed, it would be nice if we could skip the parsing phase for these types of (Void) Resources, but for now seems to be a good temporary solution 😉

Sources/Network/Parse.swift
```diff
@@ -55,4 +56,14 @@ public enum Parse {
...
+        guard data.isEmpty else {
+            throw Error.unexpectedData
```
p4checo: not sure If we need this error, or we can simply use an `assert(data.isEmpty)` 🤔

Sources/Resource/NetworkResource.swift
p4checo: also, perhaps we could create a `VoidNetworkReource: NetworkResource` for these cases, to avoid "polluting" the `NetworkResource` (and check for it on the `switch` as well) 😉

Sources/Network/Parse.swift
danieladias: I though it would be better to return an error like the other parsers. It shouldn't happen but as an API I think this is more correct than an assert.

Sources/Network/Parse.swift
danieladias: Going to ignore this one for now since it seems we are going to break Resources soon and this could probably be discussed then.

Sources/Network/Parse.swift
danieladias: Fixed

Sources/Resource/NetworkResource.swift
danieladias: I wouldn't consider this polluting since it is something that can happen in any network resource but can split.

danieladias: Changed to static var

Sources/Network/Parse.swift
danieladias: Will ignore this for now. (see above)

PR #112, opened 2018-04-02
JSON Mappable - Mappable all the things

portellaa:
We were missing this Mappable helpers in the JSON class 😄

### DIFF
--- a/Sources/Network/JSON.swift
+++ b/Sources/Network/JSON.swift
@@ -196,20 +196,27 @@ public enum JSON {
     ///   - json: The JSON dictionary.
     ///   - parseAPIError: The API error parsing closure.
     /// - Returns: The value of type `T` associated to the given attribute's key.
-    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
+    /// - Throws: An error of type `JSON.Error`, or a `Swift.Error` produced either by `parseAPIError` or the parsing function.
     public static func parseMappableAttribute<T: Mappable>(
         _ type: T.Type,
         key: JSON.AttributeKey,
         json: JSON.Dictionary,
+        where predicate: ParsePredicateClosure<T>? = nil,
         parseAPIError: ParseAPIErrorClosure? = nil) throws -> T {
 
         guard let anyValue = json[key] else {
             throw parseAPIError?(json) ?? Error.missingAttribute(key, json: json)
         }
 
-        let anyObject: Any = try parseValue(rawValue: anyValue, key: key, json: json, where: nil, parseAPIError: parseAPIError)
+        let any: Any = try parseValue(rawValue: anyValue, key: key, json: json, where: nil, parseAPIError: parseAPIError)
 
-        return try T.model(from: anyObject)
+        let mappableObject: T = try T.model(from: any)
+
+        guard predicate?(mappableObject) ?? true else {
+            throw Error.unexpectedAttributeValue(key, json: json)
+        }
+
+        return mappableObject
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Network/JSON.swift
```diff
@@ -186,6 +186,61 @@ public enum JSON {
...
+    /// the value can be used to instantiate a Mappable model. Additionally, an optional `parseAPIError` closure can be supplied
+    /// so that if a parsing step fails an attempt is made to extract a domain specific error and throw it.
+    ///
+    /// - Parameters:
+    ///   - type: The type of the attribute to parse.
+    ///   - key: The JSON attribute key to parse.
+    ///   - json: The JSON dictionary.
+    ///   - parseAPIError: The API error parsing closure.
+    /// - Returns: The value of type `T` associated to the given attribute's key.
+    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
```
p4checo: This is no longer accurate, since the `try T.model(from: anyObject)` call can throw any error
(same applies for the other calls below)

Sources/Network/JSON.swift
```diff
@@ -186,6 +186,61 @@ public enum JSON {
...
+        _ type: T.Type,
+        key: JSON.AttributeKey,
+        json: JSON.Dictionary,
+        parseAPIError: ParseAPIErrorClosure? = nil) throws -> T {
+
+        guard let anyValue = json[key] else {
+            throw parseAPIError?(json) ?? Error.missingAttribute(key, json: json)
+        }
+
+        let anyObject: Any = try parseValue(rawValue: anyValue, key: key, json: json, where: nil, parseAPIError: parseAPIError)
```
p4checo: since this is an `Any`, perhaps call this `let any = ...` ? 💄 
(same applies for the other places below)

Sources/Network/JSON.swift
```diff
@@ -186,6 +186,61 @@ public enum JSON {
...
+    ///   - key: The JSON attribute key to parse.
+    ///   - json: The JSON dictionary.
+    ///   - parseAPIError: The API error parsing closure.
+    /// - Returns: The value of type `T` associated to the given attribute's key.
+    /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
+    public static func parseMappableAttribute<T: Mappable>(
+        _ type: T.Type,
+        key: JSON.AttributeKey,
+        json: JSON.Dictionary,
+        parseAPIError: ParseAPIErrorClosure? = nil) throws -> T {
```
p4checo: could we also add the `where predicate: ParsePredicateClosure<T>? = nil` parameter?
I think it can come useful in some scenarios 😉 
(same applies for other calls below)

Sources/Network/JSON.swift
```diff
@@ -361,7 +361,7 @@ public enum JSON {
     ///   - parseAPIError: The API error parsing closure.
     /// - Returns: The value of type `T` associated to the given attribute's key.
     /// - Throws: An error of type `JSON.Error`, or a domain specific `Swift.Error` produced by `parseAPIError`.
-    private static func parseValue<T>(rawValue: Any,
+    internal static func parseValue<T>(rawValue: Any,
```
p4checo: We can perhaps simply omit the modifier, or do you think it's worth it?
Additionally, please align the parameters correctly 👮 💄

portellaa: ups 😇 and yes we can. i thought xcode would recommend to remove the modifier, but it didn't 🙄

PR #114, opened 2018-04-16
Add `Atomic` ☢️ and `Lock` 🔒

p4checo:
- Copied and modified `Atomic` and `Lock` from `ReactiveSwift` 🙏

- Converted the existing `Atomic` UT's and added some on `Lock`.

- Cleaned up test target project settings.


p4checo:
😱 🙉 😆

PR #115, opened 2018-04-17
Fix `DiskMemoryPersistenceStack` issues

p4checo:
- Fixed integer overflow when updating `usedMemorySize` since `NSCache`
eviction is made synchronously when setting a new object if the cache
is full (evict another object) or too small for the new object (evict
the object being set).

- Fixed data race when accessing `usedMemorySize`, since memory cache
set and eviction can come from multiple queues/threads (i.e. caller).

- Fixed `usedDiskSize` not being correctly updated when the file
already existed, because `createFile` overwrites any existing file with
the same path.

- Fixed disk cache eviction sorting to evict the less recently accessed
files, and not the other way around.

- Removed `URL.swift` extension, which only contained `resourceValues`,
since it was only used by `DiskMemoryPersistenceStack` and omitted
useful information on the failure cases (e.g. when files didn't exist).


Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -49,32 +45,45 @@ public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
...
     private let cache = NSCache<NSString, NSData>()
 
     private let configuration: Configuration
 
-    weak var delegate: DiskMemoryPersistenceStackDelegate?
-
     private var diskCacheEnabled: Bool = false
 
-    private let readOperationQueue: OperationQueue = {
+    let readOperationQueue: OperationQueue = {
```
portellaa: Why make those public?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -167,12 +177,14 @@ public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
             let path = self.diskPath(for: key)
 
             guard let fileData = self.fileManager.contents(atPath: path) else {
-                return completion { throw Persistence.Error.noObjectForKey }
+                return completion { [weak self] in
+                    self?.persistencePerformanceMetrics?.endReadDisk()
```
portellaa: do we begin readDisk? 😱

Tests/AlicerceTests/Persistence/DiskMemoryPersistenceTestCase.swift
```diff
@@ -31,29 +31,15 @@ final class DiskMemoryPersistenceTestCase: XCTestCase {
...
-                Thread.sleep(forTimeInterval: 5)
-                if times > 0 {
-                    checkAndRetry(times: times - 1, execute: execute)
-                }
-            }
-
-            execute(retry)
-        }
+        // wait for all operations (including the eviction ones) to finish
+        diskMemPersistence.writeOperationQueue.waitUntilAllOperationsAreFinished()
```
portellaa: i think we discussed at the time this was done to not expose properties just to make them usable by the tests. something that i don't agree and as far as i remember you neither.
well but ok 🤷‍♂️
that't why there was the 🔨 in the tests to validate to check if all the files have been written.

anyway, this is much nice way to validate, although it's breaking some "rules" (at least for me 😛 )

p4checo: It was mainly because of that test, since it was really slow and a bit weird, to be honest 😆

As for exposing these as `internal`, I know it's not "ideal", but also think that the impact should be minimal, since they won't be exposed to the framework's clients.

If you really are uncomfortable with this, we can try using a different approach to be able to inform whether all work has been completed without exposing the queues.

portellaa:
go go go ✅

PR #127, opened 2018-05-01
Router and TreeRouter improvements

p4checo:
#### Improve `Router` and `TreeRouter` API’s

- Allow multiple `RouteHandler`'s per `TreeRouter`

Since `TreeRouter` was generic on `RouteHandler` and `T` (because of the
completion closure having a `T`), only a single `RouteHandler`
implementation could be used.

To enable multiple handlers on a single `TreeRouter` (simplifying it to
`TreeRouter<T>`, a new type erasing `AnyRouteHandler<T>` was created
to wrap any handler implementation using a `T`.

- Improved `TreeRouter` API

  + Added `@discardableResult` to `unregister`
  + Changed `completion` on `route` API to `handleCompletion`, to be
clearer as to what it means (i.e. it's only run when a match is
found and the handler executes)

- Removed private queue from `TreeRouter` and replaced with an `Atomic`

- Removed `register` and `unregister` from `Router` API

#### Unnest Errors in Route types

Because the `Route.Tree` and `TreeRouter` are generic, their nested
`Error`'s also were, which made it less flexible to handle, because
each generic specialization had in practice its own `Error` type.

#### Add and reorganize `Route.X` conformances

- Added some missing `CustomStringConvertible` and
`CustomDebugStringConvertible` conformances on `Route` subtypes.

- Reorganized protocol conformances to be on extensions.

### DIFF
--- a/Sources/DeepLinking/Route+Component.swift
+++ b/Sources/DeepLinking/Route+Component.swift
@@ -73,7 +73,7 @@ extension Route.Component: Hashable {
         }
     }
 
-    public static func ==(lhs: Route.Component, rhs: Route.Component) -> Bool {
+    public static func == (lhs: Route.Component, rhs: Route.Component) -> Bool {
         switch (lhs, rhs) {
         case (.empty, .empty): return true
         case let (.constant(lhsString), .constant(rhsString)): return lhsString == rhsString
@@ -136,7 +136,7 @@ extension Route.Component.Key: Hashable {
         }
     }
 
-    public static func ==(lhs: Route.Component.Key, rhs: Route.Component.Key) -> Bool {
+    public static func == (lhs: Route.Component.Key, rhs: Route.Component.Key) -> Bool {
         switch (lhs, rhs) {
         case (.empty, .empty), (.variable, .variable): return true
         case let (.constant(lhsString), .constant(rhsString)): return lhsString == rhsString
--- a/Sources/DeepLinking/Route+Component.swift
+++ b/Sources/DeepLinking/Route+Component.swift
@@ -10,7 +10,7 @@ import Foundation
 
 public extension Route {
 
-    public enum Component: ExpressibleByStringLiteral, CustomStringConvertible, CustomDebugStringConvertible, Hashable {
+    public enum Component {
         case empty // for default handlers, e.g.: /home
         case constant(String)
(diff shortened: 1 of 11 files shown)


Sources/DeepLinking/Route+Component.swift
```diff
@@ -51,82 +51,115 @@ public extension Route {
...
+        switch self {
+        case .empty: return "\(type(of: self).empty)".hashValue
+        case let .constant(value): return "\(type(of: self).constant)(\(value))".hashValue
+        case let .variable(parameter): return "\(type(of: self).variable)(\(String(describing: parameter)))".hashValue
         }
+    }
 
-        public init(unicodeScalarLiteral value: String) {
-            self.init(component: value)
+    public static func ==(lhs: Route.Component, rhs: Route.Component) -> Bool {
```
renatorodrigues: I tried to check the style guide but it's not clear. I believe operators should be surrounded by a single whitespace when defining them. (For instance, SwiftLint enforces such rule by default). It makes it clearer and easier to perceive the operator format.

Examples (taken from SwiftLint):

```
func <| (lhs: Int, rhs: Int) -> Int {}
func <|< <A>(lhs: A, rhs: A) -> A {}
```

instead of

```
func <|(lhs: Int, rhs: Int) -> Int {}
func <|<<A>(lhs: A, rhs: A) -> A {}
```

p4checo: I think we don't have that rule specified on the style guide, indeed.

However, I think that on declaration the operator can be seen more like a function (which it is), and not as an operator (in the sense that it's being applied to operands that is 😛). 

So by looking at it as a function, in a way it makes sense for me to not have a whitespace before the `(`.

Still, perhaps we can start that discussion on our style guide repo, as like you said SwiftLint doesn't differentiate between the two cases, and when we integrate it we will have to add exceptions on all operator declaration points, if we do it like this (no whitespace before the `(`. We currently have to do that on one of our projects, and it sucks 😓.

p4checo: Ok, so after thinking about this for a while, I can see how it might be better if we have a whitespace before the `(`. This is especially evident when we gave generic operators, where the `<` can create some visual confusion with the operator itself.

Will update the operators 😉

PR #128, opened 2018-05-02
Add `GeneratableObjectCache`

filipe-lemos:
## Changes

Added `GeneratableObjectCache` that allows obtaining cached "heavy-to-generate" objects from a lightweight key object implementing `GeneratableObjectKey` protocol.

For example, using `DateFormatter` objects which are relatively heavy to compute, we can use `String`s as keys to generate/retrieve cached instances of `DateFormatter`. Key types can be as simple or complex as the specific project requires.


Tests/AlicerceTests/Utils/GeneratableObjectCacheTestCase.swift
```diff
@@ -0,0 +1,149 @@
...
+
+        func generate() -> DateFormatter {
+            let formatter = DateFormatter()
+            formatter.dateFormat = dateFormat
+            formatter.timeZone = TimeZone(identifier: "UTC")
+            return formatter
+        }
+    }
+
+    private enum ComplexDateFormatterKey: GeneratableObjectKey {
```
p4checo: nice touch adding these cases/examples! 😉👍

Tests/AlicerceTests/Utils/GeneratableObjectCacheTestCase.swift
```diff
@@ -0,0 +1,149 @@
...
+
+        func generate() -> DateFormatter {
+            let formatter = DateFormatter()
+            switch self {
+                case .utc: formatter.timeZone = TimeZone(identifier: "UTC")
+                case .local: break // local device timezone
+            }
+            switch self {
+            case .local(let dateFormat),
+                 .utc(let dateFormat):
```
p4checo: this case could be on the same line, but it's 💄

Tests/AlicerceTests/Utils/GeneratableObjectCacheTestCase.swift
```diff
@@ -0,0 +1,149 @@
...
+//  Alicerce
+//
+//  Created by Filipe Lemos on 02/05/2018.
+//  Copyright © 2018 Mindera. All rights reserved.
+//
+
+import XCTest
+@testable import Alicerce
+
+extension String: GeneratableObjectKey {
```
renatorodrigues: Although this is a unit test, I'm still not sure about having a `generate` method in the `String` namespace that generates a `DateFormatter` using as format the string contents. Since I don't believe such practice should be done in production code, having has an example might give the wrong impression as well.

Sources/Utils/GeneratableObjectCache.swift
```diff
@@ -0,0 +1,31 @@
...
+//  GeneratableObjectCache.swift
+//  Alicerce
+//
+//  Created by Filipe Lemos on 02/05/2018.
+//  Copyright © 2018 Mindera. All rights reserved.
+//
+
+import Foundation
+
+public protocol GeneratableObjectKey: Hashable {
```
renatorodrigues: Not sure about this naming – `GeneratableObjectKey` and `GeneratableObjectCache`. `Generatable` is not a very common name and in the Swift ecosystem the common term used is `Generator`.

What do you think @filipe-lemos  and @p4checo?

Tests/AlicerceTests/Utils/GeneratableObjectCacheTestCase.swift
filipe-lemos: I think you're right. The original method was `dateFormatter` and it made sense to have it as an extension of `String`, but not as much now.

I share the opinion we should remove that test as it doesn't bring much value as it is and might be interpreted as a bad example.

Sources/Utils/GeneratableObjectCache.swift
filipe-lemos: Personally I dislike the term `Generator` used in this context. While Swift doesn't use the term generator related to iterators, I still think those are deeply tied in general.

Issue #124, opened 2018-04-30
Integrate SwiftLint


p4checo:
We should integrate SwiftLint in the project (Xcode build phase and CI), enforcing as many rules as possible from our [style guide](https://github.com/Mindera/swift-style-guide) 💪

beloso:
I can pick this up ☝️

p4checo:
Great! Not sure if it helps, but we're using a `.swiftlint.yml` on one of our current projects, that might be a good starting point:

```yaml
colon:
  apply_to_dictionaries: false
force_unwrapping: error
force_try: warning
disabled_rules:
    - closure_parameter_position
    - todo
    - unused_optional_binding
    - nesting
    - weak_delegate
    - large_tuple
    - type_name
    - generic_type_name
    - cyclomatic_complexity
    - identifier_name
opt_in_rules:
    - empty_count
    - closure_spacing
    - closure_end_indentation
    - weak_delegate
    - force_unwrapping
    - nimble_operator
    - object_literal
type_body_length:
  warning: 300
  error: 400
function_body_length:
  error: 150
number_separator:
  minimum_length: 5
empty_count: warning
excluded:
    - Carthage
    - Pods/
reporter: "xcode"
```

One important thing to note is that `SwiftLint` has a bug on the colon rule ([SwiftLint/#1901](https://github.com/realm/SwiftLint/issues/1901)), and we're still using `0.22.0` because of that 😓

beloso:
Cool, I will take that into account.

I have a question though. The best way (IMO) to integrate SwiftLint into the project would be with a Podfile indicating that we need SwiftLint. That way we could use the same path across all machines that will build the framework. 

Since Cocoapods are not currently integrated in Alicerce, do you think this is a good compromise?

p4checo:
I think it should be fine. We'll have a bit more setup when using the framework, but it might be worth it.

Some things though:

1 - What are the other alternatives?
2 - If we follow the CocoaPods path, we have to make sure Carthage builds still work fine (i.e. swiftlint must be properly configured when ran on CI environment)

Meanwhile, let's see what others have to say 😉

beloso:
The options are:

- Home Brew
    - We would need to ensure that the `$PATH` is properly setup across all machines that will build the framework
    - I am not sure we can force Home Brew to install version 0.22.0

- CocoaPods
    - The `$PATH` would be agnostic.

Others are using `Mint` and a downloadable pkg.

[...]

PR #133, opened 2018-05-12
Add SwiftLint to the project

beloso:
+ ~Add Mint installation to Travis~
+ ~Use Mint package manager to integrate SwiftLint at version 0.22.0~
+ Added `Brewfile` to install Swiftlint from homebrew
+ Use base SwiftLint rules from #124
+ Fix some warnings raised by SwiftLint
+ Add SwiftLint target to run checks when building


beloso:
I am unsure that Mint will install in Travis, does anyone know a better way to do this?

p4checo:
Just looked at your setup, and you created a new target for SwiftLint on the project, which I haven't understood why 🤔

In my opinion, SwiftLint should be a (run script) build phase of the main target. That way any triggered warnings and errors will make the build fail, if I'm not mistaken.

We could perhaps also benefit of using `swiftlint autocorrect`, so that some minor fixes are done "automagically" by swiftlint.

Sources/Logging/Log+FileLogDestination.swift
```diff
@@ -74,10 +72,10 @@ public extension Log {
                     let fileHandle = try FileHandle(forWritingTo: self.fileURL)
 
                     fileHandle.seekToEndOfFile()
-                    fileHandle.write("\n".data(using: .utf8)! + formattedLogItemData)
+                    let newlineData = "\n".data(using: .utf8).require(hint: "Unable to convert string to UTF8")
```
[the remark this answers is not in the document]
beloso: Will do. I was unsure what route we wanted to follow with these errors.

Sources/Persistence/CoreData/SiblingContextCoreDataStack.swift
```diff
@@ -30,10 +30,10 @@ public final class SiblingContextCoreDataStack: CoreDataStack {
                 shouldMigrateStoreAutomatically: Bool = true,
                 shouldInferMappingModelAutomatically: Bool = true,
                 storeLoadCompletionHandler: @escaping (Any, Error?) -> Void = { (store, error) in
-                    if let error = error {
-                        fatalError("💥: Failed to load persistent store \(store)! Error: \(error)")
-                    }
-                },
+        if let error = error {
```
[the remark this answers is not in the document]
beloso: `let` should be on the next line?

beloso:
There are still quite a few warnings about line_length. Should we disable that rule?

I can make SwiftLint a build phase of the main target. But I think that having it on a separate target, that if it fails will fail the main target, i.e. the Main Target depends on the SwiftLint target. But I can move it to the main target as well.

I ran `autocorrect` before pushing this.

Alicerce.xcodeproj/project.pbxproj
```diff
@@ -1012,6 +1040,23 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "[SwiftLint] Lint files";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if which mint >/dev/null; then\n    mint run realm/SwiftLint@0.22.0 swiftlint\nfi\n";
```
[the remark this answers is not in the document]
beloso: I need to test installing Alicerce using carthage. But I was under the impression that Carthage builds the target, and if it wouldn't find mint installed it would fail the build for people using Carthage, which I think is not desirable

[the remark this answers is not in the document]
beloso: Making that change and trying out Carthage on a test project I got an error. Without this, it worked without an issue. Therefore I think we should not force Mint to be installed.

Alicerce.xcodeproj/project.pbxproj
```diff
@@ -967,6 +989,11 @@
 						LastSwiftMigration = 0910;
 						ProvisioningStyle = Manual;
 					};
+					F12FB1B920A6E88F00AA3A52 = {
+						CreatedOnToolsVersion = 9.3;
+						DevelopmentTeam = 74L475LDQT;
```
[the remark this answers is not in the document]
beloso: I removed the DevelopmentTeam from the branch.

beloso:
I am having trouble getting mint to work in Travis. Mint is taking too long to install swiftlint. And Travis is timing out the build. I don't know what can be done in this case.

Sources/Logging/Log+FileLogDestination.swift
```diff
@@ -74,10 +72,11 @@ public extension Log {
                     let fileHandle = try FileHandle(forWritingTo: self.fileURL)
 
                     fileHandle.seekToEndOfFile()
+                    //swiftlint:disable force_unwrapping
```
p4checo: Change this to `//swiftlint:disable:next force_unwrapping`, so that we don't have to re-enable it afterwards 😛

Sources/Persistence/CoreData/CoreDataEntity.swift
```diff
@@ -49,7 +49,9 @@ public extension CoreDataEntity where Self: NSManagedObject {
 
     @available(iOS, obsoleted: 10, message: "use `NSManagedObject`'s `init(context:)` instead")
     init(in context: NSManagedObjectContext) {
+        //swiftlint:disable force_unwrapping
```
p4checo: same as above

Sources/Persistence/CoreData/SiblingContextCoreDataStack.swift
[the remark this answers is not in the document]
p4checo: this is a default parameter value for the closure parameter, and it's being indented as if it were the main function's body code.

if this was done by swiftlint's auto correct, I think perhaps we should try to add an exception for it to avoid applying this indentation here.

Alicerce.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
p4checo: Ok, didn't think about that, makes sense 😅

Even though I would prefer for it to fail, I understand that because of Carthage we can't force it to be installed. 

Its just that any contributor without it installed will not be "aware" that Mint's missing until it breaks after opening up a PR (assuming the changes would cause a linter issue).

Possibly a nice middle ground would be to fire a warning if Mint was missing, but don't know if it's easy to do that from a shell script.

p4checo:
> There are still quite a few warnings about `line_length`. Should we disable that rule?

I don't think we should disable that rule. I do think we should fix the warnings though 😆 . However, I can do that on a separate PR, if you prefer. 

Regarding "line breaking rules" , I believe that there isn't yet a properly defined ruleset on our style guide. On the most common case of a long method declaration/invocation, we simply break the parameters into multiple lines (aligned by parameter name), but on other, more complex situations there isn't a clear rule, unfortunately. That's another nice discussion topic for us to raise at our [swift-style-guide](https://github.com/Mindera/swift-style-guide).

> I ran `autocorrect` before pushing this.

What I meant was that we could also run it automatically on the build, so that some minor issues can get fixed without user intervention.

> I am having trouble getting mint to work in Travis. Mint is taking too long to install swiftlint. And Travis is timing out the build. I don't know what can be done in this case.

Well that sucks 😓. I tried it myself locally, and running it the first time on Xcode, it took a long time to compile SwiftLint, threw a bunch of warnings and errors and failed to build SwiftLint itself lol 🤦‍♂️.

Then I tried to run it directly on the terminal (`mint run realm/SwiftLint@0.22.0 swiftlint`), it also took a while to compile (~2 mins), but then it worked fine on Xcode.

Perhaps we can try to install it beforehand on Travis, using `mint install realm/SwiftLint@0.22.0` after the `brew bundle` call?

Sources/Persistence/CoreData/SiblingContextCoreDataStack.swift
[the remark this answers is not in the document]
beloso: It's not directly related with SwiftLint. It was complaining about parameters not being vertically aligned. I had to put the closure on the same line as the parameter name and xcode indenter made the indentation like this. I will revert the change.

Sources/Logging/Log+FileLogDestination.swift
beloso: Didn't knew this one. Thank you. Will use it

p4checo: I had already mentioned it on the previous comment 😆 
You also have the `:this` variant, which is used inline

beloso:
> I don't think we should disable that rule. I do think we should fix the warnings though 😆 . However, I can do that on a separate PR, if you prefer.
>
> Regarding "line breaking rules" , I believe that there isn't yet a properly defined ruleset on our style guide. On the most common case of a long method declaration/invocation, we simply break the parameters into multiple lines (aligned by parameter name), but on other, more complex situations there isn't a clear rule, unfortunately. That's another nice discussion topic for us to raise at our swift-style-guide.

I will give it a try at resolving the warnings. If I can't do them I will ask you to look at them.
There is a warning about a file being too long. In that one I might require your help. I don't feel confident enough to "break" existing classes.

> What I meant was that we could also run it automatically on the build, so that some minor issues can get fixed without user intervention.

I thought about that too. But then I wasn't sure if everyone wants their code "changing" every time they hit "build". It's easy to add though. I can add it in my next commit.

> Perhaps we can try to install it beforehand on Travis, using mint install realm/SwiftLint@0.22.0 after the brew bundle call?

I will give it a try.

Alicerce.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
beloso: I will look into throwing a warning about Mint not being installed.

p4checo:
Regarding file length, for now we can add a `// swiftlint:disable:this file_length` on the last line of that file 😉 

As for `autocorrect`, let's see what others have to say. It can add some overhead on each build, but also save some time by automatically fixing some minor issues. But it's like you said, it's easily added. 👍

Alicerce.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
portellaa: I think that is not possible to throw an warning on the from a shell script, they are basically based on exit status codes (simple bash).
Perhaps send an echo with a warning? i know, it's not ideal and perhaps most of the developers would miss that, but, just my 2 cents.

`echo "⚠️⚠️☢️ ... 😂"`

.travis.yml
```diff
@@ -4,6 +4,7 @@ osx_image: xcode9.3
 xcode_project: Alicerce.xcodeproj
 
 before_install:
+  - brew bundle
```
portellaa: don't you need install? perhaps for travis you don't, just asking 😊

filipe-lemos:
We're updating the SwiftLint configuration file for our project which includes some opt-in rules that might help the overall code style consistency in Alicerce too, such as #135 header comments.

```
excluded:
    - Carthage
    - Pods/
reporter: "xcode"
disabled_rules:
    - closure_parameter_position
    - todo
    - unused_optional_binding
    - nesting
    - weak_delegate
    - large_tuple
    - type_name
    - generic_type_name
    - cyclomatic_complexity
    - identifier_name
opt_in_rules:
    - empty_count
    - closure_spacing
    - closure_end_indentation
    - force_unwrapping
    - nimble_operator
    - object_literal
    - explicit_init
    - fatal_error_message
    - file_header
    - first_where
    - implicit_return
    - multiline_parameters
    - unneeded_parentheses_in_closure_argument
    - vertical_parameter_alignment_on_call

colon:
  apply_to_dictionaries: false
force_unwrapping: error
force_try: warning
type_body_length:
  warning: 300
  error: 400
function_body_length:
  error: 150
number_separator:
  minimum_length: 5
empty_count: warning
superfluous_disable_command: warning
file_header:
  required_pattern: |
                    \/\/
                    \/\/  .*?\.swift
                    \/\/  Alicerce
                    \/\/
                    \/\/  Created by .*? on \d{2}\/\d{2}\/\d{4}\.
                    \/\/  Copyright © \d{4} Mindera\. All rights reserved\.
                    \/\/
```

Alicerce.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
renatorodrigues: It's possible as I used something similar before. You can check an example [here](http://jeffreysambells.com/2013/01/31/generate-xcode-warnings-from-todo-comments).

.travis.yml
beloso: OSX Travis images have homebrew pre-installed :)

beloso:
I've fixed more warnings from SwiftLint.
I still haven't managed to fix all of the `line_length` warnings. They are mostly coming from clojure parameters, which are incredibly long. @p4checo would you like to have a go at them? (Sorry for passing this up, I am out of ideas for those long clojures).

There are also some warnings coming from the `file_header` rule. I am waiting for conclusion of discussion at #135 in order to proceed.

We still have the problem of Travis builds timing out while installing/building SwiftLint via Mint.
Ideally we would bundle a bin file in the repo. But I could not find any option for Mint to do that.
We can go with CocoaPods path and commit the Pods folder. But that's too big of a structural change to be taken lightly.

Any thoughts?

p4checo:
@beloso :trollface: 

giphy-downsized

beloso:
Love you too @p4checo ❤️ 

You are right, I haven't been paying attention to this lately. 
Last time we talked I said I would give a try to Danger. 
I will they to accomplish this this weekend.

Sources/Logging/Loggers/Log+DefaultLogger.swift
```diff
@@ -28,18 +28,19 @@ extension Log {
...
         /// The logger's registered modules.
-        private let _modules = Atomic<[Module : Log.Level]>([:])
+        private let _modules = Atomic<[Module: Log.Level]>([:])
 
         /// The logger's error callback closure, invoked whenever any of its destinations fails an operation.
-        public var onError: ((LogDestination, Error) -> ())? = { destination, error in
-            print("💥[Alicerce.Log]: Failed to perform operation in destination '\(destination.id)' with error: \(error)")
+        public var onError: ((LogDestination, Error) -> Void)? = { destination, error in
+            print("💥[Alicerce.Log]: Failed to perform operation in destination " +
+                "'\(destination.id)' with error: \(error)")
```
p4checo: Even though we don't have a rule for this, we typically break the string "as late as possible", except if breaking it sooner makes it easier to read™. 💅

Sources/Logging/Loggers/ModuleLogger.swift
```diff
@@ -27,6 +27,7 @@ public protocol ModuleLogger: Logger {
     ///   - file: The file from where the log was invoked.
     ///   - line: The line from where the log was invoked.
     ///   - function: The function from where the log was invoked.
+    // swiftlint:disable:next function_parameter_count
```
p4checo: I believe this will mess with the documentation above it. Does Alt-clicking still work fine?

Sources/Persistence/CoreData/CoreDataEntity.swift
p4checo: `//swiftlint:disable:next force_unwrapping` ? 💅

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -508,3 +512,4 @@ private extension NSError {
         return domain == NSCocoaErrorDomain && (code == NSFileReadNoSuchFileError || code == NSFileNoSuchFileError)
     }
 }
+// swiftlint:enable file_length
```
p4checo: this isn't necessary because SwiftLint works on a per-file basis  😉 

Also, on our current project we are also only disabling it at the beginning of the file

Sources/Resource/ResourceRetry.swift
```diff
@@ -118,6 +118,7 @@ public enum ResourceRetry {
         ///   - payload: The received remote payload.
         ///   - response: The received response
         /// - Returns: The action to take.
+        // swiftlint:disable:next function_parameter_count
```
p4checo: same as above (documentation)

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -11,8 +11,8 @@ public enum NetworkPersistableStoreError: Swift.Error {
 }
 
 public class NetworkPersistableStore<Network: NetworkStack, Persistence: PersistenceStack>: NetworkStore
-where Network.Remote == Data, Network.Request == URLRequest, Network.Response == URLResponse,
-      Persistence.Remote == Data {
+    where Network.Remote == Data, Network.Request == URLRequest, Network.Response == URLResponse,
```
p4checo: here we are starting to see the weird line breaking cases (and possibilities) that come with generics and type declarations 😅 

Personally I'm a fan of aligning the `where` at the same level as `func`, but once again this is not defined anywhere on our style guide AFAIK.

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -35,16 +35,15 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
 
     @discardableResult
     public func fetch<R>(resource: R, completion: @escaping NetworkStoreCompletionClosure<R.Local, E>) -> Cancelable
-    where R: NetworkResource & PersistableResource & StrategyFetchResource & RetryableResource,
-          R.Remote == Remote, R.Request == Request, R.Response == Response
-    {
+        where R: NetworkResource & PersistableResource & StrategyFetchResource & RetryableResource,
```
p4checo: nesting the `where` like this has the side effect below, of also nesting the body once more "tab", which IMO is 💩

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -54,45 +53,46 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
...
+                        },
+                        cacheMiss: { completion(.failure(.network(networkError))) },
+                        failure: { completion(.failure(.multiple([networkError, $0]))) })
+                })
+
+            return cancelable
     }
 
     @discardableResult
+    // swiftlint:disable:next function_body_length
```
p4checo: does this have to be between `@discardableResult` and `private func`? 🤢

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -171,53 +171,55 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
                             resource: R,
                             cancelable: CancelableBag,
                             completion: @escaping NetworkStoreCompletionClosure<R.Local, E>)
-    where R: NetworkResource & PersistableResource, R.Remote == Remote {
+        where R: NetworkResource & PersistableResource, R.Remote == Remote {
```
p4checo: same as above (and more cases below)

Sources/Analytics/Analytics.swift
```diff
@@ -4,7 +4,7 @@ import Foundation
 public enum Analytics {
 
     /// A typealias representing a dictionary of analytics parameters.
-    public typealias Parameters<Key: AnalyticsParameterKey> = [Key : Any]
+    public typealias Parameters<Key: AnalyticsParameterKey> = [Key: Any]
```
p4checo: even though this style guide "rule" might change, I though that this was fixed in the latest SwiftLint version? 🤔

Sources/Stores/NetworkPersistableStore.swift
beloso: I am reverting these changes (the ones I find at least). 
I don't disagree with you. Xcode's built in "formatter" was the one you indented the `where` clause.

p4checo: yeah, Xcode can be quite "opinionated" on indentation 😝

Sources/Stores/NetworkPersistableStore.swift
beloso: I do agree. Reverting as mentioned before

Sources/Stores/NetworkPersistableStore.swift
beloso: Yeah it needs to be either in the line icmmeadetly before or after the line causing the problem. I changed this to the line after, but I don't think it got any better.

Sources/Analytics/Analytics.swift
beloso: Apparently not 😞 . I believed it was fixed but it was because I ran `seiftlint autocorrect` which corrected these occurrences, leading me to think that the exception to the rule was working 😞 .

I did revert the dictionaries to have one space around the colons in dictionaries.

Sources/Logging/Loggers/ModuleLogger.swift
beloso: > doesn't it work if we put it before the documentation?

That was my first attempt. The warning persisted. But I found a way to not use the `previous` here. I used `this`. I used the `disable:this` right next to the first parameter declaration. Since all parameters are "aligned" it goes around this issue a bit.

It not looks like this:

```
    /// - Parameters:
    ///   - module: The module from which the message originated.
    ///   - level: The severity level of the message.
    ///   - message: The log message.
    ///   - file: The file from where the log was invoked.
    ///   - line: The line from where the log was invoked.
    ///   - function: The function from where the log was invoked.
    func log(module: Module, // swiftlint:disable:this function_parameter_count
             level: Log.Level,
             message: @autoclosure () -> String,
             file: StaticString,
             line: UInt,
             function: StaticString)
```

Not "perfect", it's the best I can do with the current state of the tools

.swiftlint.yml
```diff
@@ -0,0 +1,46 @@
...
+    - todo
+    - unused_optional_binding
+    - nesting
+    - weak_delegate
+    - large_tuple
+    - type_name
+    - generic_type_name
+    - cyclomatic_complexity
+    - identifier_name
+    - colon
```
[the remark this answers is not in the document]
beloso: If I enable this rule we get 70 warnings mainly due to the dictionaries being like `[Key : Value]`
To fix these errors we would need to make them like `[Key: Value]`, since the exception does not work.

I can fix the warnings, but it would imply that we change our style guide.
We can also "humanly" ignore the warnings.

What do ou think?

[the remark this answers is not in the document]
p4checo: So essentially the rule is still broken 😅 

Fine by me, I prefer to keep it disabled and "manually" check on CR's for now.

PR #134, opened 2018-05-16
Rename `GeneratableObjectCache` to `BuilderCache`

filipe-lemos:
* Rename `GeneratableObjectCache` to `BuilderCache`
* Add `Builder` protocol and compose `BuilderKey` from it
* Rename `value` method to `object` to more closely follow cache naming conventions
* Add `evict` and `evictAll` methods to cache
* Add unit tests for new evict methods


p4checo:
@portellaa, I think you're mixing some concepts and seeing this Cache as something that it isn't (at least it wasn't thought to be). This cache is to be used for storing expensive objects that live for quite some time (e.g. `DateFormatter`'s, `NSRegularExpression`'s, etc), and not on a more "typical" scenario of a network cache. Another important difference is that in this Cache, the Keys generate the objects they'll map to. Please take a look at the tests to see some nice (practical) use cases.

The simplification/refactor you mention of the `DiskMemoryPersistenceStack` into composable chainable caches/stores is something different, and I agree that we should definitely do it 😜. However, as you well said, I think it's best if we discuss that on another place.

PS: It's great to have you back here 😘

portellaa:
Yes i know the purpose of this, but i was thinking on making it more generic and remove the "Generatable" responsibility from it, although this isn't ready for asynchronous calls, although we could offer this mechanisms, but well ok 👍 😊 
I understand the purpose 😉 

PS: Sorry for my absence 🌹

PR #135, opened 2018-05-18
Update header comments to match the overall used style.

renatorodrigues:
Some files throughout the project had inconsistencies in their header comment (filename, missing extension or package name).


Tests/AlicerceTests/Persistence/CoreData/CoreDataStackSetupTests.swift
```diff
@@ -1,6 +1,6 @@
 //
 //  CoreDataStackSetupTests.swift
-//  mobile_mrp_mrporter
+//  Alicerce
 //
 //  Created by André Pacheco Neves on 14/03/2017.
 //  Copyright © 2017 net-a-porter. All rights reserved.
```
mschuetz72: The (c) doesn't seem quite right in this file, and others... 😊

p4checo: Oops, well spotted! 😅

renatorodrigues: I didn't notice that... 😱 Thanks!

Tests/AlicerceTests/Logging/LogTests.swift
```diff
@@ -1,6 +1,6 @@
 //
 //  LogTests.swift
-//  AlicerceTests
+//  Alicerce
```
p4checo: Shouldn't these still be `AlicerceTests`, since they're on the test target?

I know most of them aren't, but perhaps we should fix this the other way around, because when we create a new file and add it to the test target, it assumes the `AlicerceTests` by default.

The fact that most of them are using `Alicerce` is probably because they were created incorrectly by initially being added to the main target 😓

renatorodrigues: I have mixed feelings about this. Xcode uses the target name when creating new files but, on the other hand, with SPM where we don't have such notion of a target, I'm unsure we need to follow this convention. I've taken a look at some popular repositories and found out that they use the main target or no specification at all.

What are your thoughts? Should we append 'Tests'?

p4checo: No strong opinion regarding this. Just mentioned because in a way we're going against Xcode's default behaviour. Nonetheless, that isn't by any means a blocker 😉 

If we decide to use `Alicerce` everywhere, we just have to remember to always fix it whenever we create a new file on the test target.

Thankfully, I think we will be able to enforce this easily using SwiftLint, right @filipe-lemos? 😇

portellaa:
What about just remove the header from all the files?

renatorodrigues:
I agree with @portellaa. It's a conventional approach that I think it makes sense.

p4checo:
Fine by me. We can even [configure that on the Xcode project](https://stackoverflow.com/a/46345626/1921751) so it won't even generate it again ✨ Haven't tested it though.

My only concern is the copyright notice. Any one knows if there's any implication from a license/legal perspective?  Or is your idea to use a simpler header (with just the filename and copyright, for instance)?

renatorodrigues:
We should only need the license file at the root of the repository.

Good summary of best practices [here](https://opensource.stackexchange.com/questions/4534/how-to-apply-mit-license-in-the-github-repo-for-fully-free-project):

> 3. License header
> 
> No a license header is not necessary, at least for MIT [...]. Some other licenses encourage you to add such a header but that is not the case [...].
> 
> The list of licenses I was mentioning earlier includes instructions on how to apply each license. For MIT a[...], it is just to copy the text of the license in a LICENSE file at the root of your repository.

> 4. Copyright header
> 
> A list of each copyright owner in each file is not necessary as well, and even projects which do maintain such headers generally forget to update them at every change. [...]

renatorodrigues:
Regarding the default file header comment, I have some bad news.

It's possible to use a custom comment, but neither Xcode 9.3 or 10 allows no comment at all.

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>FILEHEADER</key>
	<string></string>
</dict>
</plist>
```

generates:

screenshot 2018-06-12 11 02 52

It's a known issue. You can read more about it [here](https://oleb.net/blog/2017/07/xcode-9-text-macros/).

renatorodrigues:
Since it's impossible for now to have no file header at all, I will leave the copyright message. However, we should delete these two first lines whenever creating new files

p4checo:
Well it sucks that we can't remove it completely. 😓 

As for the pre-commit hook, we can definitely do that. However, if we add this rule on the linter we can also catch it quite easily on builds and on CI, so it might not be worth the hassle.

However if most people see benefit in adding the hook, I don't see why not 😄

renatorodrigues:
I believe enforcing this on the linter is the best solution.

portellaa:
Well, the solution that @renatorodrigues proposed, with just the copyright info seems nice 😃

p4checo:
Agreed (linter and message). Go go go 🚀

PR #139, opened 2018-05-28
Add missing `public init`

filipe-lemos:


portellaa:
annndd it strikes again 😂

PR #141, opened 2018-05-29
`Equatable` and `Hashable` revision

IvoPintodaSilva:
Equatable and Hashable can now be [auto synthesized](https://github.com/apple/swift-evolution/blob/master/proposals/0185-synthesize-equatable-hashable.md).

### DIFF
--- a/Sources/DeepLinking/Route+Component.swift
+++ b/Sources/DeepLinking/Route+Component.swift
@@ -53,7 +53,7 @@ public extension Route {
 
         // MARK: Key
 
-        public enum Key {
+        public enum Key: Hashable {
             case empty
             case constant(String)
             case variable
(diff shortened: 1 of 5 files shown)


Sources/DeepLinking/Route+Component.swift
```diff
@@ -53,7 +53,7 @@ public extension Route {
 
         // MARK: Key
 
-        public enum Key {
+        public enum Key: Hashable, Equatable {
```
jwelton: I believe `Hashable` inherits from `Equatable`. I think you should be able to just conform to `Hashable` here, which gives you `Equatable` conformance for free.

Sources/Network/Pinning/ServerTrustEvaluator+Configuration.swift
```diff
@@ -96,13 +96,9 @@ extension ServerTrustEvaluator {
...
 }
 
-extension ServerTrustEvaluator.PinningPolicy: Hashable {
+extension ServerTrustEvaluator.PinningPolicy: Equatable {
 
     public static func == (lhs: ServerTrustEvaluator.PinningPolicy, rhs: ServerTrustEvaluator.PinningPolicy) -> Bool {
         return lhs.domainName == rhs.domainName
     }
-
-    public var hashValue: Int {
```
p4checo: I think we also need this custom implementation so that only the domain is considered for the hash, which is the desired behaviour (uniquing factor) when adding policies to a `Set`

Sources/DeepLinking/Route+Component.swift
IvoPintodaSilva: Yep, that's right! Thanks! 🙂

Sources/Network/Pinning/ServerTrustEvaluator+Configuration.swift
```diff
@@ -98,11 +98,11 @@ extension ServerTrustEvaluator {
 
 extension ServerTrustEvaluator.PinningPolicy: Hashable {
 
-    public static func == (lhs: ServerTrustEvaluator.PinningPolicy, rhs: ServerTrustEvaluator.PinningPolicy) -> Bool {
```
p4checo: the changes on this file can simply be reset 😜

IvoPintodaSilva:
Done @p4checo, my bad! 😇

p4checo:
no worries! 😉 go for it 🚀

PR #145, opened 2018-06-19
Logging improvements (Modules, Metadata, others)

### LINKED ISSUE #126: Add modules/tags to Logging
Currently, our Logging infrastructure doesn't support modules/tags, which are useful in some scenarios, like:

 - grouping logs by tag
 - filtering logs by tag
 - enable/disable logs by tag

The base API should use `String` for the modules/tags, and then users can create their own extensions using a more type safe API (e.g. using `enum MyModule: String`).

p4checo:
## Changes

- Add support for Modules (tags), to provide even more control over
what gets logged. Only registered modules' messages whose log level is
above the module's minimum log level are logged. Log destinations' log
level filtering is preserved.

  + Created a `LogModule` protocol, which must be a `RawRepresentable`
conforming type, with a `RawValue == String`.

  + `Logger` now has an associatedtype Module: LogModule`, modules
can be registered/unregistered, and logging API's receive a new
`Module` parameter.

  +  Moved `Log`'s `Logger` conformance to a new `DefaultLogger`
generic class, which contains the base implementation.

- Add support for custom Metadata, to enable setting arbitrary data on
to the logs (e.g. user id, email, device id, etc). This extra
information can be very handy on its own, and also to correlate logs
between providers, for instance.

  + Created new `setMetadata` and `removeMetadata` API's on `Logger`
and `LogDestination`.

- Add current queue label and timestamp to `Log.Item`

  + Created an extension on `DispatchQueue` to retrieve the current
queue's label.

  + The timestamp of each `Log.Item` is now set on its creation, rather
than when being formatted, which is more correct because of async
logging.

- Added an `associatedtype Output` on `LogItemFormatter` to allow more
flexibility on the output (e.g. output directly as `Data`).

- Removed `LogItemFormatter` requirement from `LogDestination`.

- Removed `Queue` parameter from `ConsoleLogDestination` because most
basic printing functions are already thread safe (e.g. `print`,
`NSLog`, `os_log`). In the unlikely the function isn't thread safe, one
can easily wrap the function in an asynchronous call inside the closure.

- Optimized `Atomic` locking times by using `value` instead of
`withValue`.

- Made basic `log` function a part of `Logger`, and moved level
specific `log()` functions to protocol extension on `Logger`.

- Reorganized some `Log`'s nested types.

- Made `Log.Level` and `Log.Item` conform to `Codable`.

- Made `JSONLogItemFormatter` use a `JSONEncoder`.

## Issues

Closes #126


Sources/Logging/Protocols/LogItemFormatter.swift
```diff
@@ -7,7 +7,8 @@
 //
 
 public protocol LogItemFormatter {
-    func format(logItem: Log.Item) -> String
-}
 
+    associatedtype Output
 
+    func format(logItem: Log.Item) throws -> Output
```
renatorodrigues: 💄 Why not just call it `item`? The context is clear enough.

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
```diff
@@ -0,0 +1,69 @@
...
+
+        private let output: OutputClosure
+        private let logMetadata: LogMetadataClosure?
+
+        // MARK: - Lifecycle
+
+        public init(formatter: ItemFormatter,
+                    minLevel: Level = .error,
+                    output: @escaping OutputClosure = { (level, message) in print(message) },
+                    logMetadata: LogMetadataClosure? = nil) {
```
renatorodrigues: Isn't `metadata` enough?

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public func setMetadata(_ metadata: [AnyHashable : Any], failure: @escaping (Error) -> Void) {
+
+            guard let (level, message) = logMetadata?(metadata), !message.isEmpty else { return }
+
+            output(level, message)
+        }
+
+        public func removeMetadata(forKeys keys: [AnyHashable], failure: @escaping (Error) -> Void) {
```
renatorodrigues: Should we add a comment stating this method does nothing to prevent the consumers from being confused with it?

Sources/Extensions/DispatchQueue.swift
```diff
@@ -0,0 +1,19 @@
+//
```
renatorodrigues: Don't forget to remove the headers.

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
```diff
@@ -0,0 +1,69 @@
...
+
+            self.minLevel = minLevel
+            self.formatter = formatter
+            self.output = output
+            self.logMetadata = logMetadata
+        }
+
+        // MARK: - Public methods
+
+        public func write(item: Log.Item, failure: @escaping (Error) -> Void) {
```
renatorodrigues: What do you think about specialize the error type as well? Like, `class ConsoleLogDestination<ItemFormatter: LogItemFormatter, Error: ConsoleLogDestinationError>`? Or would it cause a pain in the 🍑?

Sources/Logging/Destinations/Log+FileLogDestination.swift
```diff
@@ -0,0 +1,118 @@
...
+
+    public class FileLogDestination<ItemFormatter: LogItemFormatter>: LogDestination
+    where ItemFormatter.Output == Data {
+
+        public typealias LogMetadataClosure = ([AnyHashable : Any]) -> Data
+
+        public let formatter: ItemFormatter
+
+        public let minLevel: Level
+        public lazy var id: String = "\(type(of: self))_\(fileURL.absoluteString)"
```
renatorodrigues: `private(set) lazy var`?

Sources/Logging/Destinations/Log+FileLogDestination.swift
```diff
@@ -0,0 +1,118 @@
...
+                                                                               metadataData,
+                                                                               error))
+                }
+            }
+        }
+
+        public func removeMetadata(forKeys keys: [AnyHashable], failure: @escaping (Error) -> Void) {
+        }
+
+        private func write(logData: Data) throws {
```
renatorodrigues: 💄 Simply `data`?

Sources/Logging/Destinations/Log+FileLogDestination.swift
```diff
@@ -0,0 +1,118 @@
...
+            guard self.fileManager.fileExists(atPath: fileURL.path) else {
```
renatorodrigues: No need for `self`.

Sources/Logging/Log.swift
```diff
@@ -8,136 +8,137 @@
...
-        }
-    }
+        public private(set) var destinations = Atomic<[LogDestination]>([])
+        public private(set) var modules = Atomic<[Module : Log.Level]>([:])
 
-    public private(set) var destinations = Atomic<[LogDestination]>([])
-    public var errorClosure: ((LogDestination, Item, Swift.Error) -> ())? = { destination, item, error in
-        print("💥[Alicerce.Log]: Failed to log item \(item) to destination '\(destination.id)' with error: \(error)")
-    }
+        public var errorClosure: ((LogDestination, Swift.Error) -> ())? = { destination, error in
```
renatorodrigues: 💄 `onError` reads better than `errorClosure`. The type already tells us it's expecting a closure, while the name itself tells us it's a handler called on situations of failure.

Sources/Logging/Log.swift
```diff
@@ -8,136 +8,137 @@
...
+        }
 
-    public init() {}
+        public init() {}
 
-    // MARK: - Destination Management
+        // MARK: - Destination Management
 
-    public func register(_ destination: LogDestination) throws {
+        public func registerDestination(_ destination: LogDestination) throws {
```
renatorodrigues: 😢

Sources/Logging/Log.swift
```diff
@@ -8,136 +8,137 @@
...
+                }
+            }
+        }
 
-        log(level: .info, message: message, file: file, function: function, line: line)
-    }
+        // MARK: - Logging
 
-    public func warning(_ message: @autoclosure () -> String,
+        public func log(module: Module,
```
renatorodrigues: Should we provide a method to log messages without a module?

Sources/Logging/Protocols/LogDestination.swift
```diff
@@ -13,10 +13,12 @@ public protocol LogDestination: class {
     typealias ID = String
 
     var minLevel: Log.Level { get }
-    var formatter: LogItemFormatter { get }
     var id: ID { get }
 
-    func write(item: Log.Item, failure: @escaping (Error) -> ())
+    func write(item: Log.Item, failure: @escaping (Error) -> Void)
```
renatorodrigues: 💄 `onFailure`?

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
p4checo: `metadata` might be interpreted as the metadata itself, whereas the objective is for this to be the closure that "logs the metadata"

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
p4checo: you mean a comment in the body, for instance? (i.e. not documentation, right?)

Sources/Logging/Log.swift
p4checo: Well spotted! Completely agree 😉

Sources/Logging/Log.swift
p4checo: Yeah, I know 😢 

However I felt the final API wouldn't perhaps be very clear if we used two overrides for `register(_:)` and `unregister(_:)`

Do you think it's worth it? (i.e. makes the API as clear as it is now, from the point of calling)

Sources/Logging/Protocols/LogDestination.swift
p4checo: As above, agreed 😄 

My thought process when naming this was from the perspective of `completion` / `failure`, but `onFailure` in a way names the "event" itself better.

Sources/Logging/Log.swift
p4checo: Also thought about it, but then I decided that one could simply have a single module on the app.
It has the downside that we have to use it on every call, though, so I see why it can be cumbersome 😅. One could always make some helper calls in an extension to abstract that away, but it's still not ideal.

That being said, even though I prefer using logging modules across my apps, I can see benefit in continuing to allow the previous (simpler) logging pattern.

Providing this API without a `Module` parameter can be achieved in multiple ways:

1. add another `log` API (requiring no `Module`), performing no module filtering on the `Logger`.
2. modify the existing `log` API to receive a `Module?`, and ignore module filtering on the `Logger` when `nil`.
3. make `LogModule` have a `default` or `none` value, use that on the new `log` API to send to the existing one, and ignore module filtering on the `Logger` for this "special" module.

Option 1 will result in two distinct API's, preserve backwards compatibility (allowing a smooth transition), and avoid extra complexity on the existing `log` implementation (filtering by module).

Options 2 and 3 will still require one to implement a new `log` API without module (internally sending in `nil` or the default value to the ), and will result in more complexity in the `log` implementation (and `LogModule`). This new helper `log` API will also provide backwards compatibility.

Thoughts? Other alternatives?

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
p4checo: It's doable, but I don't see any clear benefit in doing so. 🤔 Can you please elaborate?

The error handling on the `Logger` has to be generic (i.e. work on `Error` alone), and each destination should have (and already has) its own error type according to its implementation  (if needed).

Alternatively, we could create a generic `LogDestinationError` that all destinations would use, with errors for the main operations, like:

```swift
enum LogDestinationError: Error {
    case itemFormat(Log.Item, Error?)
    case write(Log.Item, Error?)
    case setMetadata([AnyHashable : Any], Error?)
    case removeMetadata([AnyHashable], Error?)
    case other(Error?)
}
```

But once again, I don't see any practical value in this because (for now) we're using a `throws` based API and as such on the `Logger` we wouldn't have type safe errors 😓

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
renatorodrigues: Agree with you, although I think `log metadata` reads as a verb instead of a callback. If I read the signature, and see `logMetadata`, my first reaction is to assume it expects a boolean. But on the other hand, I agree it should be named purely `metadata`

I can't come up with a better naming though. Obviously, this is not a blocker; just me being picky. We have the `output` closure. Could we call this one `metadataOutput`?

Sources/Logging/Log.swift
renatorodrigues: With the addition of the modules, this change makes sense indeed. Nevermind!

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
renatorodrigues: I meant documentation.

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
renatorodrigues: The advantages do not justify the disadvantages of this, especially because this failure closure has a really little impact on the logging system. Nevermind. 🙇‍♂️

Sources/Logging/Log.swift
renatorodrigues: Meanwhile, you went already for the option number 2. Good call and good job. ✨

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
p4checo: Not a very big fan of `metadataOutput`, since it doesn't output the metadata directly. It creates the input (level and message) for the `output` closure to log 😝 

Perhaps `metadataMessage`? If not, we can just leave it as is for now, I guess 😇 

Meanwhile, I'll add documentation to the API's so (hopefully) it should be clearer to the users.

Sources/Logging/Destinations/Log+ConsoleLogDestination.swift
p4checo: Will do

Sources/Logging/Destinations/Log+FileLogDestination.swift
```diff
@@ -43,6 +78,9 @@ public extension Log {
 
         // MARK: - Public Methods
 
+        /// Clears the destiantion's log file, by deleting the file. The file is recreated on write if needed.
```
renatorodrigues: Typo: `destiantion`

Sources/Logging/Destinations/Log+FileLogDestination.swift
```diff
@@ -2,30 +2,65 @@ import Foundation
...
 public extension Log {
 
+
+    /// An error produced by a `FileLogDestination`.
+    ///
+    /// - clearFailed: The log file couldn't be cleared (removed).
+    /// - itemFormatFailed: Failed to format a log item failed.
+    /// - itemWriteFailed: Writing a log item to file failed.
+    /// - metadataWriteFailed: Writing metadata message to file failed.
     public enum FileLogDestinationError: Error {
```
renatorodrigues: I believe we should add a comment for each case instead of putting them all together in the enum documentation. It's what Apple does, and it allows one to consult the description of each case (from anywhere in the code). Also, it should prove to be easier to maintain in the long term whenever we rename/remove cases.

Sources/Logging/Log+DefaultLogger.swift
```diff
@@ -0,0 +1,204 @@
...
+                        file: StaticString = #file,
+                        line: UInt = #line,
+                        function: StaticString = #function) {
+
+            // skip module checks for `nil` modules
+            if let module = module {
+                guard
+                    let moduleMinLevel = modules.value[module],
+                    level.isAbove(minLevel: moduleMinLevel)
+                    else { return }
```
renatorodrigues: Wrong indentation.

p4checo: Well spotted, damn Xcode and its opinionated auto formatting 😓

Sources/Logging/Destinations/Log+FileLogDestination.swift
p4checo: You're right. I did this because Xcode auto generates it like this, but to use alt click on the cases themselves we have to define it on each one. 😉

PR #147, opened 2018-07-01
Split `Logger` protocol and add type safe log metadata keys

p4checo:
# Changes

- Split `Logger` into `Logger`, `ModuleLogger` and `MetadataLogger` protocols, where the latter has a new `MetadataKey` associated type, for more type safe metadata keys.

- Split `LogDestination` into `LogDestination` and `MetadataLogDestination` protocols, where the latter has a new `MetadataKey` associated type, for more type safe metadata keys.

- Updated `Log.DefaultLogger`, `Log.ConsoleLogDestination` and `Log.FileLogDestination` to use the new protocols.

- Created new type erased `AnyMetadataLogDestination`, to enable storing multiple log destinations with the same `MetadataKey` on `DefaultLogger`.

- Reorganized Logging folder structure.

- Fixed some Log related documentation.


Sources/Logging/Destinations/AnyMetadataLogDestination.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    /// The type erased destination's `removeMetadata` method, stored as a closure.
+    private let _removeMetadata: ([MetadataKey], @escaping (Error) -> Void) -> Void
+
+    /// Creates a type erased instance of a log destination that outputs logs to the console and sets/unsets custom
+    /// logging metadata.
+    ///
+    /// - Parameters:
+    ///   - d: The log destination to erase the type from.
+    public init<D: MetadataLogDestination>(_ d: D) where D.MetadataKey == MetadataKey {
```
renatorodrigues: `destination` instead of `d` ✨

PR #149, opened 2018-07-04
Add some missing UT's + some UT clean ups

p4checo:
# Changes 

- Added `String` UT's.

- Added `HTTP` UT's, and removed `RawRepresentable` conformance from `HTTP.StatusCode`.

- Added `URLRequest` UT.

- Added `Box` and `VarBox` UT's.

- Improved CV/TV Views and Cells' UT's.

- Removed `ApplicationMode`, since it only makes sense on the main target, or when using CocoaPods (where the compile flags can be configured). It's also not extensible, so it's usefulness is very limited.

- Removed usage of `DispatchQueue.global(qos: .default)` in `MockPersistenceStack` API's, since it isn't necessary and might introduce concurrency issues on UT's. Also added extra `xInvokedClosure` closures to `MockPersistenceStack`,  for completeness.


Sources/Extensions/Optional.swift
```diff
@@ -1,30 +1,30 @@
...
-     *
-     *  This method will either return the value that this optional contains, or trigger
-     *  a `preconditionFailure` with an error message containing debug information.
-     *
-     *  - parameter hint: Optionally pass a hint that will get included in any error
-     *                    message generated in case nil was found.
-     *
-     *  - return: The value this optional contains.
-     */
+// Credits to John Sundell https://github.com/JohnSundell/Require/
```
renatorodrigues: 💄Splitting this into two lines is more readable:

```
// Credits to John Sundell 
// https://github.com/JohnSundell/Require/
```

p4checo: changed it for consistency with other places, but can change to two lines 💅

PR #151, opened 2018-07-10
Add Performance Metrics improvements 📈

p4checo:
# Changes 

- Made `PerformanceMetrics` a namespace (case-less enum).

- Renamed the `PerformanceMetrics` default implementation to `MultiTracker`, where the sub trackers are injected on `init`.

- Updated `PerformanceMetricsTracker`:

  + Renamed `begin`/`end` API's to `start`/`stop`

  + Updated `start` to return a `Token<Tag>` to uniquely identify a metric instance, and `stop` to receive a `token` instead of the `identifier`, to allow multiple concurrent metrics with the same identifier to be collected.

  + Moved `measure` and `measureAsync` into the protocol, and refactored them into two `measure` variants, with different `stop` mechanics. They both have a default protocol extension implementation.

- Removed `PersistencePerformanceMetrics`, and created a `PersistencePerformanceMetricsTracker` protocol that extends `PerformanceMetricsTracker`.

- Updated `DiskMemoryPersistenceStack` to use a `PersistencePerformanceMetricsTracker`.

- Removed `ResourceParsePerformanceMetrics` and created a `NetworkStorePerformanceMetricsTracker` protocol that extends `PerformanceMetricsTracker`.

- Updated `NetworkPersistableStore` to use a `NetworkStorePerformanceMetricsTracker`.

- Created a `XCTAssertDumpsEqual` helper function to avoid crashing when using `assertDumpsEqual` on UT's, and instead fail the test.

### DIFF
--- a/Sources/PerformanceMetrics/PerformanceMetrics.swift
+++ b/Sources/PerformanceMetrics/PerformanceMetrics.swift
@@ -23,11 +23,12 @@ public enum PerformanceMetrics {
         private let tokenizer = Tokenizer<Tag>()
 
         /// The tracker's token dictionary, containing the mapping between internal and sub trackers' tokens.
-        private let tokens = Atomic<[Token<Tag>: [Token<Tag>]]>([:])
+        private let tokens = Atomic<[Token<Tag> : [Token<Tag>]]>([:])
 
         /// Creates a new performance metrics multi trcker instance, with the specified sub trackers.
         ///
-        /// - Parameter trackers: The sub trackers to forward performance measuring events to.
+        /// - Parameters:
+        ///   -trackers: The sub trackers to forward performance measuring events to.
         public init(trackers: [PerformanceMetricsTracker]) {
             assert(trackers.isEmpty == false, "🙅‍♂️: trackers shouldn't be empty, since it renders this tracker useless!")
 
@@ -37,15 +38,15 @@ public enum PerformanceMetrics {
         // MARK - PerformanceMetricsTracker
 
         /// Starts measuring the execution time of a specific code block identified by a particular identifier, by
-        /// propagating it to all registered sub tracker
+        /// propagating it to all registered sub tracker.
         ///
-        /// - Parameter identifier: The metric's identifier, to group multiple metrics on the provider.
+        /// - Parameters:
+        ///   - identifier: The metric's identifier, to group multiple metrics on the provider.
         /// - Returns: A token identifying this particular metric instance.
         public func start(with identifier: Identifier) -> Token<Tag> {
 
             let token = tokenizer.next
-
-            let subTokens = trackers.map { $0.start(with: identifier) }
+            let subTokens = startTrackers(with: identifier)
 
             tokens.modify { $0[token] = subTokens }
 
@@ -65,9 +66,7 @@ public enum PerformanceMetrics {
                 return
             }
 
-            assert(subTokens.count == trackers.count, "😱: number of sub tokens and sub trackers must match!")
-
-            zip(trackers, subTokens).forEach { (tracker, token) in tracker.stop(with: token, metadata: metadata) }
+            stopTrackers(with: subTokens, metadata: metadata)
         }
 
         /// Measures a given closure's execution time once it returns or throws (i.e. *synchronously*). An optional
@@ -85,10 +84,8 @@ public enum PerformanceMetrics {
                                metadata: Metadata? = nil,
                                execute: () throws -> T) rethrows -> T {
 
-            let subTokens = trackers.map { $0.start(with: identifier) }
-            defer {
-                zip(trackers, subTokens).forEach { (tracker, token) in tracker.stop(with: token, metadata: metadata) }
-            }
+            let subTokens = startTrackers(with: identifier)
+            defer { stopTrackers(with: subTokens, metadata: metadata) }
 
             let measureResult = try execute()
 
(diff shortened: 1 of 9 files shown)


Sources/PerformanceMetrics/PerformanceMetrics.swift
```diff
@@ -1,52 +1,124 @@
...
+        /// - Returns: The closure's return value, if any.
+        /// - Throws: The closure's thrown error, if any.
+        @discardableResult
+        public func measure<T>(with identifier: Identifier,
+                               execute: (_ stop: (_ metadata: Metadata?) -> Void) throws -> T) rethrows -> T {
+
+            let subTokens = trackers.map { $0.start(with: identifier) }
+
+            return try execute { [trackers] metadata in
+                zip(trackers, subTokens).forEach { (tracker, token) in tracker.stop(with: token, metadata: metadata) }
```
renatorodrigues: We repeat this `zip/forEach/stop` 3 times in this file. Should we move this logic to a private function?

Sources/PerformanceMetrics/PerformanceMetrics.swift
```diff
@@ -1,52 +1,124 @@
...
+        ///   - stop: The closure to be invoked by `execute` to stop measuring the execution time, along with any
+        /// additional metric metadata.
+        ///   - metadata: The metric's metadata dictionary.
+        /// - Returns: The closure's return value, if any.
+        /// - Throws: The closure's thrown error, if any.
+        @discardableResult
+        public func measure<T>(with identifier: Identifier,
+                               execute: (_ stop: (_ metadata: Metadata?) -> Void) throws -> T) rethrows -> T {
+
+            let subTokens = trackers.map { $0.start(with: identifier) }
```
renatorodrigues: We repeat this logic to create the subtokens 3 times in this file. Is it worth to move this to a function?

Sources/Persistence/PersistencePerformanceMetricsTracker.swift
```diff
@@ -0,0 +1,159 @@
...
+public protocol PersistencePerformanceMetricsTracker: PerformanceMetricsTracker {
+
+    /// A closure to be used when stopping measuring memory reads/writes.
+    typealias MemoryAccessStopClosure<E: Error> = (_ result: Result<(blobSize: UInt64, memorySize: UInt64), E>) -> Void
+
+    /// A closure to be used when stopping measuring disk reads/writes.
+    typealias DiskAccessStopClosure<E: Error> = (_ result: Result<(blobSize: UInt64, diskSize: UInt64), E>) -> Void
+
+    /// The metadata key used for the used memory.
+    var usedMemoryMetadataKey: PerformanceMetrics.Metadata.Key { get }
```
renatorodrigues: If you add a `typealias Metadata = PerformanceMetrics.Metadata` to `PerformanceMetricsTracker`, you don't need to keep repeating so much code, making it easier to read.

Sources/Persistence/PersistencePerformanceMetricsTracker.swift
```diff
@@ -0,0 +1,159 @@
...
+    var usedDiskMetadataKey: PerformanceMetrics.Metadata.Key { get }
+
+    /// The metadata key used for the blob size being read/written.
+    var blobSizeMetadataKey: PerformanceMetrics.Metadata.Key { get }
+
+    /// The metadata key used for the error.
+    var errorMetadataKey: PerformanceMetrics.Metadata.Key { get }
+
+    /// The metric identifier for a memory read operation.
+    var memoryReadIdentifier: PerformanceMetricsTracker.Identifier { get }
```
renatorodrigues: You can use only `Identifier` instead of `PerformanceMetricsTracker.Identifier`.

Tests/AlicerceTests/PerformanceMetrics/PerformanceMetrics+MultiTrackerTestCase.swift
```diff
@@ -0,0 +1,170 @@
...
+        subTrackerA = nil
+        subTrackerB = nil
+
+        super.tearDown()
+    }
+
+    // start
+
+    func testStart_ShouldInvokeStartOnAllSubTrackers() {
+        let expectationA = self.expectation(description: "startA")
```
renatorodrigues: Unnecessary use of  `self`.

Tests/AlicerceTests/PerformanceMetrics/PerformanceMetrics+MultiTrackerTestCase.swift
```diff
@@ -0,0 +1,170 @@
...
+        subTrackerB.startInvoked = {
+            XCTAssertEqual($0, testIdentifier)
+            expectationB.fulfill()
+        }
+
+        let _ = tracker.start(with: testIdentifier)
+    }
+
+    func testStart_WithMultipleInvocationsForTheSameIdentifier_ShouldReturnDifferentTokens() {
+        let expectationA = self.expectation(description: "startA")
```
renatorodrigues: Unnecessary use of  `self`.

Tests/AlicerceTests/PerformanceMetrics/PerformanceMetrics+MultiTrackerTestCase.swift
```diff
@@ -0,0 +1,170 @@
...
+            expectationB.fulfill()
+        }
+
+        tracker.stop(with: testToken, metadata: testMetadata)
+    }
+
+    // measure (synchronous stop)
+
+    func testMeasure_WithSynchronousStop_ShouldInvokeStartAndStopOnAllSubTrackers() {
+        let startA = self.expectation(description: "startA")
```
renatorodrigues: Unnecessary use of  `self`.

Sources/PerformanceMetrics/PerformanceMetrics.swift
p4checo: Well spotted! Will change 😉

Sources/PerformanceMetrics/PerformanceMetrics.swift
p4checo: I didn't do it since it's such a simple logic (and copy paste is super efficient these days 😇)

Sources/Persistence/PersistencePerformanceMetricsTracker.swift
p4checo: Well spotted. We already have those `typealias`'es on `PerformanceMetricsTracker`, forgot to clean these up here. Will fix 😄

PR #153, opened 2018-07-19
Implement resource retry mechanisms ♻️

p4checo:
## Changes

- Created new `RetryableResource` protocol, to represent a resource that can be retried.

  + A default handling logic has been implemented, which combines multiple policies into an action, each time an error occurs.

- Created new `ResourceRetry` namespace, containing most types used for the resource retry logic. Most relevant ones are:

  + `ResourceRetry.Action` represents an action to take after evaluating a policy.

  + `ResourceRetry.Policy` represents a retry policy, with multiple available strategies.

  + `ResourceRetry.Policy.Rule` represents a custom retry policy rule.

- Updated `URLSessionNetworkStack` to take advantage of the new `RetryableResource` capabilities, while also deprecating the usage of `authenticator.isAuthenticationInvalid` for auth related retries.

- Removed `NetworkAuthenticator`'s `isAuthenticationInvalid` method, which essentially has been deprecated in favor of using a custom retry policy rule.

- Created new `RetryableNetworkAuthenticator` extending `NetworkAuthenticator`, which requires the implementation of a method to return a custom `RetryPolicy.Rule` for authentication related retries.

- Updated `NetworkStack`, `NetworkStore` and conforming types to also require a `RetryableResource`, and thus have a `Request` and `Response` associated types.

- Revised `PersistenceStack` APIs to use correct types.

- Revised `NetworkPersistableStore` by unnesting some types, revising some API's and `Cancelable` related logic.

- Unified `Cancelable` and conforming base types in the same place, added conformance to `URLSessionTask` and `DispatchWorkItem`.

- Created dedicated files for some Network related mocks.

### DIFF
--- a/Sources/Network/URLSessionNetworkStack.swift
+++ b/Sources/Network/URLSessionNetworkStack.swift
@@ -3,45 +3,18 @@ import Result
 
 public extension Network {
 
-    public final class CancelableTask: Cancelable {
-        private weak var task: URLSessionTask?
-
-        init(task: URLSessionTask) {
-            self.task = task
-        }
-
-        public func cancel() {
-            task?.cancel()
-        }
-    }
-
-    public final class CancelableBag: Cancelable {
-        private lazy var cancelables: [Cancelable] = []
-
-        public init() {}
-
-        public func add(cancelable: Cancelable) {
-            cancelables.append(cancelable)
-        }
-
-        public func cancel() {
-            cancelables.forEach { $0.cancel() }
-        }
-    }
-
-    public struct NoCancelable: Cancelable {
-        public func cancel() {}
-    }
-
     final class URLSessionNetworkStack: NSObject, NetworkStack, URLSessionDelegate {
 
         public typealias Remote = Data
+        public typealias Request = URLRequest
+        public typealias Response = URLResponse
 
         public typealias URLSessionDataTaskClosure = (Data?, URLResponse?, Swift.Error?) -> Void
 
         private let authenticationChallengeHandler: AuthenticationChallengeHandler?
         private let authenticator: NetworkAuthenticator?
         private let requestInterceptors: [RequestInterceptor]
+        private let retryQueue: DispatchQueue
 
         public var session: URLSession? {
             // In order to define `self` as the session's delegate while preserving dependency injection, the session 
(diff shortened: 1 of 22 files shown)


Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -219,9 +203,54 @@ public extension Network {
...
+            case (.none, _):
+                completion(.failure(.url(error)))
+            case (.retry, _):
+                let retryCancelable = fetch(resource: resource, completion: completion)
+
+                cancelableBag.add(cancelable: retryCancelable)
+            case (.retryAfter(let delay), _):
+                resource.totalRetriedDelay += delay
+
+                let fetchWorkItem = DispatchWorkItem(block: { [weak self] in
```
renatorodrigues: 💄 You can use trailing closure syntax.

Sources/Resource/ResourceRetry.swift
```diff
@@ -0,0 +1,148 @@
...
+                                 _ request: Request,
+                                 _ error: Swift.Error,
+                                 _ payload: Remote?,
+                                 _ response: Response?) -> Action
+
+        /// A backoff strategy that defines an amount of time for each retry, as well as a truncation mechanism to
+        /// limit the retries (either by number of retries, or total delay time).
+        public enum Backoff {
+
+            /// A backoff strategy that delays each retry by a constant amount of time, while a truncation rule allows.
```
renatorodrigues: "[...] while a truncation rule allows it."

Sources/Resource/RetryableResource.swift
```diff
@@ -0,0 +1,78 @@
...
+    associatedtype Request
+
+    /// A type that represents a network response.
+    associatedtype Response
+
+    /// The resource's specialized retry policy.
+    typealias RetryPolicy = ResourceRetry.Policy<Remote, Request, Response>
+
+    /// The errors that have resulted in a retry after having occurred.
+    var retriedAfterErrors: [Error] { get set }
```
renatorodrigues: It might be just me, but I find this name hard to interpret. Would `pastErrors` be enough/ better?

p4checo: We're using `previousErrors` on other API's (i.e. `ResourceRetry.Policy`), but I went with `retriedAfterErrors` on the Resource to make it clear that they are errors related with the retry "behavior". They're in also the errors that have in fact resulted in a retry after having occurred 😝.

However it the name is confusing or hard to interpret, I guess we can change it.

PR #155, opened 2018-08-28
UIImage Convenience Init - Base64 String

jwelton:
This PR just adds a convenience init to `UIImage` for initialising from base64 encoded strings. Currently this is a two step process requiring you to construct a `Data` object and then a `UIImage` object. Both of these init's are optional meaning the caller has to account for two cases of `nil`. This convenience init wraps these two optionals calls into one.

### DIFF
--- a/Tests/AlicerceTests/UIKit/UIImageTestCase.swift
+++ b/Tests/AlicerceTests/UIKit/UIImageTestCase.swift
@@ -20,4 +20,28 @@ final class UIImageTestCase: XCTestCase {
 
         XCTAssertEqual(mrMinderTemplate.renderingMode, .alwaysTemplate)
     }
+
+    func testConvenienceInit_WithBase64String_ShouldReturnUIImage() {
+        let mrMinder = imageFromFile(withName: "mr-minder", type: "png")
+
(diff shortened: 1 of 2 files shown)


Tests/AlicerceTests/UIKit/UIImageTestCase.swift
```diff
@@ -20,4 +20,28 @@ final class UIImageTestCase: XCTestCase {
...
+
+        guard let image = UIImage(base64Encoded: base64String) else {
+            XCTFail("Could not init UIImage from base64 representation")
+            return
+        }
+
+        XCTAssertEqual(UIImagePNGRepresentation(image), pngImageData)
+    }
+
+    func testConvenienceInit_WithNoneBase64String_ShouldReturnNilFromInit() {
```
filipe-lemos: Maybe also add a unit test with a valid Base64 string that isn't a valid image, such as `Ym9vbQ==`

jwelton: Good idea 👍 Added in 9a72ccca293fc35bcb8f1d3de73827bdfa607ca9

PR #156, opened 2018-08-29
Improve `DiskMemoryPersistenceStack` 📼

p4checo:
## Changes

- Revised `DiskMemoryPersistenceStack`'s concurrency to allow injecting custom `DispatchQueue`s for reading and writing, with the `readQueue` being optional so that when set to `nil` reads from disk are made *synchronously*, avoiding a context switch (queue jump).

- Updated `PersistenceStack` to use `Result` based completion handlers, and to have an associated Error type (instead of forcing `Persistence.Error` to be used).

- Removed `Persistence.Error`, since now cache misses are represented as returning a `nil` value (instead of `Error.noObjectForKey`, and we avoid an error nesting that didn't bring much value (using `Error.other`).

- Nested `DiskMemoryPersistenceStack` inside `Persistence` namespace.

- Did some tweaks on `NetworkPersistableStore` to improve error handling.

### DIFF
--- a/Sources/Persistence/DiskMemoryPersistenceStack.swift
+++ b/Sources/Persistence/DiskMemoryPersistenceStack.swift
@@ -1,458 +1,441 @@
 import UIKit
 import Result
 
-public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
+public extension Persistence {
 
-    public struct Configuration {
+    public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
 
-        /// Disk size limit in bytes.
-        let diskLimit: UInt64
+        public struct Configuration {
 
-        /// Memory size limit in bytes.
-        let memLimit: UInt64
+            /// Disk size limit in bytes.
+            let diskLimit: UInt64
 
-        /// Path where the data is persisted.
-        let path: String
+            /// Memory size limit in bytes.
+            let memLimit: UInt64
 
-        /// The manager used for file operations.
-        let fileManager: FileManager
+            /// Path where the data is persisted.
+            let path: String
 
-        /// The performance metrics tracker.
-        let performanceMetrics: PersistencePerformanceMetricsTracker?
+            /// The manager used for file operations.
+            let fileManager: FileManager
 
-        /// Set the operation Quality of Service for read and write operations
-        /// The read value should be higher than write, otherwise it will fail
-        /// Default Values:
-        ///   - read = .userInitiated
-        ///   - write = .utility
-        var qos: (read: QualityOfService, write: QualityOfService) {
-            willSet {
-                guard newValue.read.rawValue > newValue.write.rawValue else {
-                    return assertionFailure("💥 read value should be higher than write")
+            /// The performance metrics tracker.
+            let performanceMetrics: PersistencePerformanceMetricsTracker?
+
+            /// The queue to use for read operations. Set to `nil` for synchronous reads.
+            let readQueue: DispatchQueue?
+
+            /// The underlying queue to use for write operations (i.e. by the `writeOperationQueue`).
+            let writeQueue: DispatchQueue
+
+            public init(diskLimit: UInt64,
+                        memLimit: UInt64,
+                        path: String,
+                        fileManager: FileManager = .default,
+                        performanceMetrics: PersistencePerformanceMetricsTracker? = nil,
+                        readQueue: DispatchQueue?,
+                        writeQueue: DispatchQueue) {
+
+                self.diskLimit = diskLimit
+                self.memLimit = memLimit
+                self.path = path
+                self.fileManager = fileManager
+                self.performanceMetrics = performanceMetrics
+                self.readQueue = readQueue
+                self.writeQueue = writeQueue
+            }
+        }
+
+        public enum Error: Swift.Error {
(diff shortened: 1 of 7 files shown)


Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -49,35 +50,41 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
...
-            case let .failure(error):
+            case .failure(.cancelled):
+                // The fetch was cancelled, so exit right away
+                completion(.failure(.cancelled))
+                return
+
+            case let .failure(networkError):
 
                 // Check if it's cancelled
                 guard cancelable.isCancelled == false else {
```
IvoPintodaSilva: Why is it necessary to check here if fetch is canceled? Shouldn't that case be caught on the `case .failure(.cancelled)`?

p4checo: Great question!

The `.cancelled` error that we receive here (from the `networkStack`) corresponds to having received an `NSError` with domain `NSURLErrorDomain` and code `NSURLErrorCancelled`.

This error can be caused not only any the user cancelling the `URLSessionTask` (in which case this check would be redundant), but also because of TLS/SSL issues for instance.

Hope this answers your question 😄

IvoPintodaSilva: Yep, it does 😄 
Thanks!

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -171,7 +169,13 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
 
             // update persistence with new value
             if !fromCache {
-                self.persist(payload, for: resource)
+                persistenceStack.setObject(payload, for: resource.persistenceKey) {
+                    switch $0 {
+                    case .success: break
+                    case .failure(let error):
+                        print("⚠️ [Alicerce.NetworkPersistableStore]: Failed to persist value for '\(resource)': \(error)")
```
renatorodrigues: In the future, we should allow the API consumer to provide a custom logger.

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -1,458 +1,429 @@
...
-    private let usedMemorySize = Atomic<UInt64>(0) // Size in bytes
+        private let writeOperationQueue: OperationQueue = {
+            $0.name = "com.mindera.alicerce.persistence.diskmem.write.operation-queue"
+            $0.maxConcurrentOperationCount = 1
+            return $0
+        }(OperationQueue())
 
-    public init(configuration: Configuration) {
-        self.configuration = configuration
+        private var usedDiskSize: UInt64 = 0 // Size in bytes
```
renatorodrigues: `typealias Bytes = UInt64`?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -1,458 +1,429 @@
 import UIKit
 import Result
 
-public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
+public extension Persistence {
 
-    public struct Configuration {
+    public final class DiskMemoryPersistenceStack: NSObject, PersistenceStack {
```
renatorodrigues: With this namespace, we end up having `Persistence.DiskMemoryPersistenceStack`, repeating twice the `Persistence` bit. Should we rename the class to `DiskMemoryStack`, which results in `Persistence.DiskMemoryStack`?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -1,458 +1,429 @@
...
 
-                self?.setCachedData(data, for: key)
+        public func object(for key: Key, completion: @escaping ReadCompletionClosure) {
+            if let data = cachedData(for: key) {
+                return completion(.success(data))
 
-                completion { data }
-            } catch {
-                completion { throw error }
+                // TODO: Check for object on disk and cache it if not present
```
filipe-lemos: Isn't this already being done with `setCachedData` below?

Sources/Stores/NetworkPersistableStore.swift
p4checo: Agreed, we should revisit this and other (few) cases 👍

Sources/Persistence/DiskMemoryPersistenceStack.swift
p4checo: Also thought about it, but we have the same "issue" on `Network.URLSessionNetworkStack`

The thing is that it's a `PersistenceStack`, so it makes sense for it to be a `SomethingPersistenceStack`.

But yes, I agree that it gets a bit weird now with the namespace 😆

Still not convinced with `Persistence.DiskMemoryStack` though 🤔

Sources/Persistence/DiskMemoryPersistenceStack.swift
p4checo: Initially I also had to think a bit to understand what it was (and also thought it was already done), so I improved the comment a bit to make it clearer what the `TODO` really is 😉👍

PR #157, opened 2018-09-06
Remove cached value from Store on parse error 🚽

p4checo:
In some scenarios, it's possible that the cached data becomes outdated (e.g. the API and parsing logic were updated) and breaks when being parsed. Until now, these values where never cleaned up unless
explicitly evicted by the user, or automatically by the `NSCache` and/or disk eviction process (total size based only).

To address this, the cached values are now evicted whenever parsing fails (if being read from the cache).

### DIFF
--- a/Sources/Persistence/DiskMemoryPersistenceStack.swift
+++ b/Sources/Persistence/DiskMemoryPersistenceStack.swift
@@ -55,7 +55,6 @@ public extension Persistence {
             public enum FileRemovalError: Swift.Error {
                 case fileResourceValuesFetchFailed(Swift.Error)
                 case invalidFileTypeAtPath(String)
-                case fileNotFound(Swift.Error)
                 case removeFailed(Swift.Error)
             }
 
@@ -249,7 +248,7 @@ public extension Persistence {
                     try self.remove(fileAtURL: fileURL, size: fileSize)
 
                 } catch let error as NSError where error.isNoSuchFileError {
-                    return completion(.failure(.failedToRemoveFile(.fileNotFound(error))))
+                    return completion(.success(())) // file might have been evicted already
                 } catch let error as Error {
                     return completion(.failure(error))
                 } catch {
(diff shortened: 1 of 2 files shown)


Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -424,10 +423,11 @@ public extension Persistence {
                 }
 
                 filesToRemove.forEach {
-                    guard let _ = try? self.remove(fileAtURL: $0.url, size: $0.fileAttr.size) else {
-                        assertionFailure("💥 Failed to remove file with path 👉 \($0)")
-
-                        return print("💥[Alicerce.Persistence.DiskMemoryPersistenceStack]: Failed to remove file with path 👉 \($0)")
+                    do {
+                        try self.remove(fileAtURL: $0.url, size: $0.fileAttr.size)
```
filipe-lemos: Is this `self.` necessary?

Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -249,7 +248,7 @@ public extension Persistence {
                     try self.remove(fileAtURL: fileURL, size: fileSize)
 
                 } catch let error as NSError where error.isNoSuchFileError {
```
filipe-lemos: `isNoSuchFileError` is true on `NSFileReadNoSuchFileError` but there is a similar error `NSFileNoSuchFileError` and I can't tell if `NSFileManager.removeItem` might also throw it or not.

Sources/Persistence/DiskMemoryPersistenceStack.swift
p4checo: Yes, because we're inside a `DiskMemoryBlockOperation` execution block 😉

Sources/Persistence/DiskMemoryPersistenceStack.swift
p4checo: Nice catch on `NSFileNoSuchFileError`, will handle it 👍 (even though I haven't seen it in the wild occurring from the `URL.resourceValues` call (which performs a file read)).

Regarding the `NSFileManager.removeItem` call (made inside `self.remove()`), at that point we're assuming the file should exist, and we wrap any error in a `Error.failedToRemoveFile(.removeFailed(error))` which is propagated upstream.

PR #159, opened 2018-09-07
Implement remove all from PersistenceStack

p4checo:
On some use cases, it makes sense to clear the persistence stack and evict all entries (e.g. on logout).

To address this, a `removeAll` API was added to `PersistenceStack` (and `DiskMemoryPersistenceStack`), and a `clearPersistence` API was added to `NetworkPersistableStore`.

### DIFF
--- a/Sources/Persistence/DiskMemoryPersistenceStack.swift
+++ b/Sources/Persistence/DiskMemoryPersistenceStack.swift
@@ -51,6 +51,7 @@ public extension Persistence {
             case failedToGetDirectoryContents(Swift.Error)
             case failedToRemoveFile(FileRemovalError)
             case failedToCreateFile(FileCreationError)
+            case failedToRemoveAll(RemoveAllError)
 
             public enum FileRemovalError: Swift.Error {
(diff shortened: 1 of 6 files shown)


Sources/Persistence/DiskMemoryPersistenceStack.swift
```diff
@@ -432,6 +447,33 @@ public extension Persistence {
...
+                    try configuration.fileManager.removeItem(atPath: configuration.path)
+                } catch {
+                    completion(.failure(.failedToRemoveAll(.removeDirectoryFailed(error))))
+                    return
+                }
+
+                self.usedDiskSize.value = 0
+
+                do {
+                    try configuration.fileManager.createDirectory(atPath: configuration.path,
```
filipe-lemos: Maybe use `createDiskCacheDirectory()` ?

p4checo: I purposely didn't so that the returned error would be different (i.e. would be nested in `Error.failedToRemoveAll`)

PR #164, opened 2018-09-25
Remove CommonCrypto on Xcode 10, and generate a dummy one otherwise.

renatorodrigues:


Alicerce.xcodeproj/project.pbxproj
```diff
@@ -1541,19 +1541,19 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
-		0AB0562720B81D7600DC363F /* Check for /usr/include/CommonCrypto/CommonCrypto.h */ = {
+		0AB0562720B81D7600DC363F /* CommonCrypto */ = {
```
portellaa: Why are we keeping this? Xcode 9- support?

install_common_crypto.sh
```diff
@@ -0,0 +1,21 @@
...
+
+mkdir -p "${FRAMEWORK_DIR}/Modules"
+cat <<EOF > "${FRAMEWORK_DIR}/Modules/module.modulemap"
+module CommonCrypto [system] {
+    header "${SDKROOT}/usr/include/CommonCrypto/CommonCrypto.h"
+    export *
+}
+EOF
+
+ln -sf "${SDKROOT}/usr/include/CommonCrypto" "${FRAMEWORK_DIR}/Headers"
```
p4checo: Isn't this a bit "intrusive"? 🤔 
Also, if other libs do the same "hack", this will break quite easily. Or am I missing something?

Alicerce.xcodeproj/project.pbxproj
renatorodrigues: To support Xcode 9, yes.

install_common_crypto.sh
renatorodrigues: We're just linking the framework module we just created (`FRAMEWORK_DIR` points to our `CommonCrypto.framework`). It shouldn't break since we only proceed if the framework doesn't exist.

p4checo: I understand that we're pointing to our dummy module, and we only do it if it doesn't exist. My concerns are:

1. we're creating a link inside the user's `$SDKROOT` which I think is a bit intrusive.
2. any other lib making the same "hack" will overwrite whatever is at `${SDKROOT}/usr/include/CommonCrypto` to point to their own dummy module, making "our" link broken in the process, while the check for the existence of the module keeps passing.

renatorodrigues: Hm, doesn't `ln` work the other way around – `ln -s /path/to/original/ /path/to/link`?

p4checo:
@renatorodrigues 

> I don't find this an "ugly" solution when compared to having a hardcoded module map in the repository since it has no impact for whoever is using Xcode 10.

Well, any hack of this sort is hardly beautiful 😆. Just find this one more intrusive, thus more "ugly".

> I would prefer moving forward with the complete removal of this dummy module map. Fine by me merging this to our temporary branch. This hack is to support our different versions of Xcode between the CI processes and our environments – especially with OS X Mojave where without the hack, the build breaks.

The title and description of your PR don't mention any issue with OS X Mojave, only Xcode versions. If this is necessary for the project to be built with Xcode 9 in OS X Mojave, then it's different from a fix to support both Xcode 9 and Xcode 10. Please provide more information next time for those of us who are off 😜

install_common_crypto.sh
p4checo: Oops, you're right 😅 
Please ignore my "concerns"

Alicerce.xcodeproj/project.pbxproj
portellaa: nhec 😂

portellaa:
wait, i have Mojave, i don't have the hack and it works 🤔

portellaa:
Can we merge this 😁 ?

PR #166, opened 2018-10-01
Fix CommonCrypto script error with Xcode 10.

renatorodrigues:


install_common_crypto.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+COMMON_CRYPTO_DIR="${SDKROOT}/usr/include/CommonCrypto"
+if [ -f "${COMMON_CRYPTO_DIR}/module.modulemap" ]; then
+    exit 0
+fi
+
 FRAMEWORK_DIR="${BUILT_PRODUCTS_DIR}/CommonCrypto.framework"
 
 # Check if module map exists (Xcode 10). If not, generate a dummy module.
```
p4checo: This comment should now be before the previous `if` 😝 , and here it could be another one like:

`# Check if the dummy module has already been created. If not, create it.`

install_common_crypto.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+COMMON_CRYPTO_DIR="${SDKROOT}/usr/include/CommonCrypto"
+if [ -f "${COMMON_CRYPTO_DIR}/module.modulemap" ]; then
+    exit 0
```
p4checo: for consistency, maybe we should add a log here, or remove the other one below 💄

renatorodrigues: Fixed.

PR #171, opened 2018-10-17
Improve Cancelable and Network cancellation ❌

p4checo:
# Changes

- Add `isCancelable` property to `Cancelable`.

- Improve `CancelableBag` to cancel any added cancelables if already cancelled, to ensure correctness and avoid lingering work that can't be cancelled.

- Added more checks in `URLSessionNetworkStack` to avoid scheduling retry work if the cancelable has already been cancelled.

- Handle `NSURLErrorCancelled` error in `NetworkStack`'s `NetworkStore` default conformance.

- Create extension on `NSError` to check if it's a `NSURLErrorDomain` and `NSURLErrorCancelled` error.

- Cleaned up Info.plist placement and added missing one for `HostAppRequiringTests` target.


Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -242,8 +242,7 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
             switch result {
             case .success(let data):
                 success(data)
-            case .failure(.url(let error as NSError))
-            where error.domain == NSURLErrorDomain && error.code == NSURLErrorCancelled:
+            case .failure(.url(let error as NSError)) where error.isURLErrorCancelled:
```
jwelton: Just out of curiosity. Couldn't you switch on the error type instead? I.e.
``` swift
case .failure(.url(URLError.cancelled))
```

p4checo: Great point! Even though we can (and probably should) use `URLError`, I think that we cannot switch on `URLError.cancelled`, as that's only the code of the error 😓. Still, I think we can go with:

`case .failure(.url(let error as URLError)) where error.code == .cancelled:`

😉

Tests/AlicerceTests/Shared/CancelableBagTestCase.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    func testIsCancelled_WithCancelledCancelableBag_ShouldReturnTrue() {
+        XCTAssertFalse(cancelable.isCancelled)
+        cancelable.cancel()
+        XCTAssertTrue(cancelable.isCancelled)
+    }
+
+    // add
+
+    func testAdd_WithNotCancelledCancelledBag_ShouldAddCancelableToBag() {
```
filipe-lemos: Maybe use `WithNotCancelledCancelableBag` or simply `WithNotCancelledBag` as the `Bag`s in the test refer to `CancelableBag` anyway.

p4checo: I think I've messed up those names, yes 😆

PR #172, opened 2018-11-05
Swift 4.2 migration 🚀

IvoPintodaSilva:
Migrating the codebase to Swift 4.2


Alicerce.xcodeproj/project.pbxproj
```diff
@@ -1467,21 +1449,24 @@
 				TargetAttributes = {
 					0A266F801ED59DC7009CD0D7 = {
 						CreatedOnToolsVersion = 8.3.2;
-						LastSwiftMigration = 0910;
-						ProvisioningStyle = Manual;
+						DevelopmentTeam = 74L475LDQT;
```
beloso: I believe we should not be committing this.

Alicerce.podspec
```diff
@@ -26,6 +26,4 @@ Pod::Spec.new do |s|
 
     s.frameworks    = [ 'Foundation', 'UIKit', 'CoreData', 'Security' ]
     s.dependency 'Result', '~> 4.0'
-
-    s.script_phase  = { :name => 'CommonCrypto', :script => '${PODS_TARGET_SRCROOT}/install_common_crypto.sh', :execution_position => :before_compile }
```
p4checo: We should also remove the `s.preserve_paths` line, and delete the install_common_crypto.sh file 🗑

README.md
```diff
@@ -48,7 +48,7 @@ TODO
 
 ## Installation 🔧
 
-Alicerce supports iOS 9.0+ and requires Xcode 9.3+.
+Alicerce supports iOS 9.0+ and requires Xcode 10+.
```
p4checo: We should probably add a note that for Xcode 9.3 compatibility (Swift 4.0), Alicerce 0.3.0 should be used 💅

IvoPintodaSilva:
Updated with your suggestions @beloso and @p4checo 👍

Sources/UIKit/UIViewController.swift
```diff
@@ -32,8 +32,8 @@ public extension UIViewController {
     ///   - unselectedRenderMode: UIImageRenderingMode to the unselected image. Default .alwaysOriginal
```
filipe-lemos: Documentation is still referencing old types.

IvoPintodaSilva: Nice catch! 👍

PR #175, opened 2018-11-19
Propagate URLResponse on NetworkStack

IvoPintodaSilva:
Sometimes the URLResponse brings valuable information (in the headers, as an example) for the application.
As such, it's good to be able to access them when performing network operations 🙂

### DIFF
--- a/Sources/Network/Network.swift
+++ b/Sources/Network/Network.swift
@@ -5,18 +5,35 @@ public enum Network {
 
     // MARK: - TypeAlias
 
-    public typealias CompletionClosure<R> = (Result<R, Error>) -> Void
+    public typealias CompletionClosure<R> = (Result<Value<R>, Error>) -> Void
     public typealias AuthenticationCompletionClosure = (URLSession.AuthChallengeDisposition, URLCredential?) -> Void
 
-    // MARK: - Network Error
+    // MARK: - Response value
 
-    public enum Error: Swift.Error {
-        case http(code: HTTP.StatusCode, apiError: Swift.Error?)
-        case noData
-        case url(Swift.Error)
-        case badResponse
-        case authenticator(Swift.Error)
-        case retry(errors: [Swift.Error], totalDelay: ResourceRetry.Delay, retryError: ResourceRetry.Error)
+    public struct Value<R> {
+
+        let value: R
+
+        let response: URLResponse
+    }
+
+    // MARK: - Response error
+
+    public struct Error: Swift.Error {
+
+        let type: ErrorType
+
+        let response: URLResponse?
+
+        enum ErrorType {
+
+            case http(code: HTTP.StatusCode, apiError: Swift.Error?)
+            case noData
+            case url(Swift.Error)
+            case badResponse
+            case authenticator(Swift.Error)
+            case retry(errors: [Swift.Error], totalDelay: ResourceRetry.Delay, retryError: ResourceRetry.Error)
+        }
     }
 
     // MARK: - Network Configuration
--- a/Sources/Network/Network.swift
+++ b/Sources/Network/Network.swift
@@ -14,6 +14,11 @@ public enum Network {
 
         let value: R
         let response: URLResponse
+
+        public init(value: R, response: URLResponse) {
+            self.value = value
+            self.response = response
+        }
     }
 
     // MARK: - Response error
--- a/Sources/Network/Network.swift
+++ b/Sources/Network/Network.swift
@@ -25,11 +25,11 @@ public enum Network {
 
     public enum Error: Swift.Error {
 
-        case http(code: HTTP.StatusCode, apiError: Swift.Error?, response: URLResponse?)
-        case noData(response: URLResponse?)
+        case http(code: HTTP.StatusCode, apiError: Swift.Error?, response: URLResponse)
+        case noData(response: URLResponse)
         case url(Swift.Error, response: URLResponse?)
         case badResponse(response: URLResponse?)
-        case authenticator(Swift.Error, response: URLResponse?)
+        case authenticator(Swift.Error)
         case retry(errors: [Swift.Error],
             totalDelay: ResourceRetry.Delay,
             retryError: ResourceRetry.Error,
@@ -72,8 +72,8 @@ extension Network.Error {
             return response
         case let .badResponse(response):
             return response
-        case let .authenticator(_, response):
-            return response
+        case .authenticator:
+            return nil
         case let .retry(_, _, _, response):
             return response
         }
--- a/Sources/Network/Network.swift
+++ b/Sources/Network/Network.swift
@@ -12,8 +12,8 @@ public enum Network {
 
     public struct Value<R> {
 
-        let value: R
-        let response: URLResponse
+        public let value: R
+        public let response: URLResponse
 
         public init(value: R, response: URLResponse) {
             self.value = value
--- a/Sources/Network/Network.swift
+++ b/Sources/Network/Network.swift
@@ -31,9 +31,9 @@ public enum Network {
         case badResponse(response: URLResponse?)
         case authenticator(Swift.Error)
         case retry(errors: [Swift.Error],
-            totalDelay: ResourceRetry.Delay,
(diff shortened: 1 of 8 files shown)


Sources/Network/Network.swift
```diff
@@ -5,18 +5,35 @@ public enum Network {
...
-        case http(code: HTTP.StatusCode, apiError: Swift.Error?)
-        case noData
-        case url(Swift.Error)
-        case badResponse
-        case authenticator(Swift.Error)
-        case retry(errors: [Swift.Error], totalDelay: ResourceRetry.Delay, retryError: ResourceRetry.Error)
+    public struct Value<R> {
+
+        let value: R
+
```
p4checo: extra newline? 💄

Sources/Network/Network.swift
```diff
@@ -5,18 +5,35 @@ public enum Network {
...
+        let response: URLResponse
+    }
+
+    // MARK: - Response error
+
+    public struct Error: Swift.Error {
+
+        let type: ErrorType
+
```
p4checo: same as above 💅

Sources/Network/Network.swift
```diff
@@ -5,18 +5,35 @@ public enum Network {
...
+    public struct Value<R> {
+
+        let value: R
+
+        let response: URLResponse
+    }
+
+    // MARK: - Response error
+
+    public struct Error: Swift.Error {
```
p4checo: If we have to make this conform to `Swift.Error` (because of `Result`), why not simply add a `response: URLResponse?` to all the enum cases of the `Error` (now `ErrorType`)? 

We could then make a simple computed property to extract it.

Having these two error "levels" feels a bit weird...

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -136,7 +136,7 @@ public extension Network {
 
                 guard let httpResponse = response as? HTTPURLResponse else {
                     // don't retry this error since this should "never" happen
-                    completion(.failure(.badResponse))
+                    completion(.failure(Network.Error(type: .badResponse, response: nil)))
```
p4checo: we should propagate the `response` here, as it might be non nil

And if we nest it on the `Error` enum, this would then be `completion(.failure(.badResponse(response: response)))` (or even simply `completion(.failure(.badResponse(response)))`)

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -145,17 +145,19 @@ public extension Network {
 
                 switch (httpStatusCode, data) {
                 case (.success, let remoteData?):
-                    completion(.success(remoteData))
+                    completion(.success(Network.Value(value: remoteData, response: httpResponse)))
```
p4checo: we should propagate `response` and not `httpResponse` (even though it is the same instance)

(more cases below)

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -187,8 +189,7 @@ public extension Network {
                     return strongSelf.perform(request: authenticatedRequest, resource: resource, completion: completion)
 
                 case let .failure(error):
-                    completion(.failure(.authenticator(error.error)))
-
+                    completion(.failure(Network.Error(type: .authenticator(error.error), response: nil)))
```
p4checo: we should send the `response` anyway

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -181,49 +174,75 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
...
+
+        } catch let error as Parse.Error {
+            completion(.failure(.parse(error)))
+        } catch {
+            completion(.failure(.other(error)))
+        }
+    }
+
+    private func processFromCache<R>(_ payload: Remote,
+                                       resource: R,
```
p4checo: incorrectly aligned arguments 👮

Sources/Stores/NetworkPersistableStore.swift
```diff
@@ -181,49 +174,75 @@ where Network.Remote == Data, Network.Request == URLRequest, Network.Response ==
...
+                case .failure(let error):
+                    print("⚠️ [Alicerce.NetworkPersistableStore]: Failed to remove value for '\(resource)': \(error)")
                 }
             }
         } catch {
             completion(.failure(.other(error)))
         }
     }
 
+
```
p4checo: extra newline 🚓

Tests/AlicerceTests/Network/URLSessionNetworkStackTestCase.swift
```diff
@@ -546,10 +554,8 @@ final class URLSessionNetworkStackTestCase: XCTestCase {
             switch result {
             case .success:
                 XCTFail("🔥 should throw an error 🤔")
-            case let .failure(.url(receivedError as NSError)):
-                XCTAssertEqual(receivedError, mockError)
             case let .failure(error):
-                XCTFail("🔥 received unexpected error 👉 \(error) 😱")
+                XCTAssertDumpsEqual(error, Network.Error.init(type: .url(mockError), response: mockResponse))
```
p4checo: no need for explicit `.init` 😛

Sources/Network/Network.swift
IvoPintodaSilva: Yeah, it does feel weird 😕 
I'll try the other approach

Sources/Network/URLSessionNetworkStack.swift
IvoPintodaSilva: At this point the request hasn't been performed, right?

p4checo: Oops, you're right!
In that case, the `.authenticator` error shouldn't even have a `response` 😉

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -134,9 +134,14 @@ public extension Network {
                     return
                 }
 
-                guard let httpResponse = response as? HTTPURLResponse else {
+
+
```
p4checo: extra newline 🚓

Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -134,9 +134,14 @@ public extension Network {
...
+                guard
+                    let urlResponse = response,
```
p4checo: if we now do this then most of the errors no longer require an `URLResponse?`, but a `URLResponse`, right?

IvoPintodaSilva: True, only the `.badResponse` needs an optional response 😃

IvoPintodaSilva: Actually, only `.http` and `.noData` can have non-optional responses.
And the `.authenticator` error does not feature a response associated value anymore.

Sources/Network/Network.swift
```diff
@@ -5,18 +5,35 @@ public enum Network {
...
-        case badResponse
+
+        case http(code: HTTP.StatusCode, apiError: Swift.Error?, response: URLResponse)
+        case noData(response: URLResponse)
+        case url(Swift.Error, response: URLResponse?)
+        case badResponse(response: URLResponse?)
         case authenticator(Swift.Error)
-        case retry(errors: [Swift.Error], totalDelay: ResourceRetry.Delay, retryError: ResourceRetry.Error)
+        case retry(errors: [Swift.Error],
+            totalDelay: ResourceRetry.Delay,
```
p4checo: incorrect alignment 💅

Tests/AlicerceTests/Network/Mocks/MockNetworkStack.swift
```diff
@@ -42,7 +43,7 @@ final class MockNetworkStack: NetworkStack {
             if let error = self.mockError {
                 completion(.failure(error))
             } else if let data = self.mockData {
-                completion(.success(data))
+                completion(.success(Network.Value.init(value: data, response: self.mockResponse)))
```
p4checo: no need to use `.init` 💅

PR #177, opened 2018-11-28
Support for paths with spaces in the CommonCrypto script 🙈

portellaa:


portellaa:
🙈 i forgot to update my local copy, i was pointing the pod to that version. Sorry

Thanks @p4checo 🙏

PR #179, opened 2019-02-08
Make `NetworkResource` request creation async 📬

p4checo:
## Description

Until now, `Resource` fetch authentication was made via a single `NetworkAuthenticator` injected on a `URLSessionNetworkStack`, which was used to enrich the requests of the Resources being scheduled on the stack. This had some limitations:

1. Either all `Resource` fetches were authenticated, or none (depending on whether `authenticator` was set or not on the stack).

2. Since the `authenticator` was shared when set, supporting different authentication schemes (e.g. different API's) or unauthenticated fetches on the same stack required creating a complex `NetworkAuthenticator` (containing / proxying all logic), which was obviously not ideal.

To solve the above problems and give more flexibility to our stack, the `NetworkResource` keeps being responsible for creating its requests, with the difference that now it does so *asynchronously*. This allows each `NetworkResource` implementation to create (and enrich) its requests, only passing them to the stack when they are ready to be scheduled. As such, more complex request building logic like
authentication can now be done by the Resource itself (which can have its own `NetworkAuthenticator` instance, for example).

## Changes

- Removed `NetworkAuthenticator` from `Network.URLSessionNetworkStack` and `Network.Configuration`.

- Renamed `NetworkAuthenticator` to `ResourceAuthenticator` and `RetryableNetworkAuthenticator` to `RetryableRequestAuthenticator`, and refined them to contain associated types.

- Updated `NetworkResource` to have a `Request` associated type, and a new asynchronous `makeRequest()` API.

- Removed `RelativeNetworkResource` and `StaticNetworkResource`.

- Created `HTTPResourceEndpoint` to represent an HTTP endpoint, and contain default `URLRequest` building logic.

- Created `HTTPNetworkResource` to represent a Resource with a `HTTPResourceEndpoint`, and contain a default `makeRequest()` implementation.

- Created `AuthenticatedHTTPNetworkResource` that extends `HTTPNetworkResource` and has a `NetworkAuthenticator` instance, and contains a default `makeRequest()` implementation.

- Added UT's for the new types.

- Adjusted existing UT's.

- Cleaned up and unified some `MockResource`s in UT's.


Sources/Network/RequestAuthenticator.swift
```diff
@@ -0,0 +1,53 @@
...
+public protocol RequestAuthenticator: class {
+
+    /// A type that represents a network request.
+    associatedtype Request
+
+    /// A type that represents a request authentication error.
+    associatedtype Error: Swift.Error
+
+    /// The authenticator's authentication completion closure, invoked when a request's authentication finishes.
+    typealias AuthenticateCompletion = (Result<Request, Error>) -> Cancelable
```
renatorodrigues: Shouldn't we call it `AuthenticaticationCompletion`, as you described in the comment?

Sources/Resource/AuthenticatedHTTPNetworkResource.swift
```diff
@@ -0,0 +1,24 @@
...
+}
+
+extension AuthenticatedHTTPNetworkResource {
+
+    @discardableResult
+    public func makeRequest(_ completion: @escaping MakeRequestCompletion) -> Cancelable {
+
+        return authenticator.authenticate(endpoint.request) { result in
+
+            return completion(result.mapError(AnyError.init))
```
IvoPintodaSilva: Unnecessary `return`

Sources/Network/RequestAuthenticator.swift
p4checo: I guess it can. My train of thought was to name it `AuthenticateCompletion` since the method is `authenticate`, but now that I look at it again it does sound a bit odd 😅

I'm also not completely sold on it being called "completion" (same for `makeRequest`), since it returns a `Cancelable` (meaning itself returns an "operation" that can be cancelled, and should be chained with the one returned from `authenticate`). Any suggestions?

Sources/Resource/AuthenticatedHTTPNetworkResource.swift
p4checo: Wrote it like this mostly for readability, but can definitely make it more compact 😉

PR #180, opened 2019-02-10
Bump Result to 4.1.0

p4checo:
- Updated Result to 4.1.0, hopefully for the last time before `Swift.Error` 😍.

- Ignored dirty state in Result's submodule.


Cartfile
```diff
@@ -1 +1 @@
-github "antitypical/Result" ~> 4.0.0
\ No newline at end of file
+github "antitypical/Result" ~> 4.0.0
```
jwelton: Should we also use `4.1.0` here too?

p4checo: I think this follows the semantic versioning rules (similar to CocoaPods), which in this case specifies that it will update to all versions except any major version above 4

Issue #163, opened 2018-09-24
Network Documentation


dehlen:
Hey, it looks like the Network documentation is based on an old version of this framework. To me it looks like that the example `RESTResource<T: Codable>` class has to implement the `RetryableResource` protocol. I added conformance to the protocol and the following implementation which I hope is the way to go.

` var retryErrors: [Swift.Error] = []
    var totalRetriedDelay: ResourceRetry.Delay = 0
    var retryPolicies: [ResourceRetry.Policy<Data, URLRequest, URLResponse>] = []

    typealias Request = URLRequest
    typealias Response = URLResponse`
If you would confirm this I would love to make PR in order to correct the documentation.

However I am currently searching for a way to authenticate my network stack using OAuth2 (GitHub API). I think this should be fairly easy be possible by using the `NetworkAuthenticator` and `AuthenticationChallengeHandler`, but I would love some guidance since I am not quite sure at the moment on how to implement this. Could you help me out here with some information on how to add that to the network stack ?

p4checo:
hi @dehlen 👋 

Sorry for the late reply. Yes, we forgot to update our Playground with the latest changes. 😓
The snippet you mention should do the job, so feel free to open a PR!

As for your second question, I think you might be able to achieve that via a `NetworkAuthenticator` implementation. The `AuthenticationChallengeHandler` is designed to work with `URLAuthenticationChallenge` API's, which are used at a lower level of authentication (e.g. establishing an encrypted TLS session) and shouldn't be what you're looking for.

So in essence you could create an `OAuth2NetworkAuthenticator: NetworkAuthenticator` implementing the OAuth2 protocol logic, and then inject it on your `URLNetworkStack`. That way, whenever the network stack performs a fetch and requests an authentication (via `authenticate`), your authenticator would enrich the `URLRequest` with the necessary credentials, or trigger the mechanisms to fetch/refresh them and then enrich the request when possible.

If you also make your authenticator conform to `RetryableNetworkAuthenticator`, you will then have to supply a retry policy (which you set on your Resources) that will allow you to react to authentication errors (e.g. 403's), trigger internal token refresh mechanisms and obviously determine whether a request should be retried or not from an authentication perspective.

Hope this made sense! Cheers! 🍻

PR #190, opened 2019-04-14
Fix Network playground 🎳

p4checo:
As mentioned in #163, the Network playground has been broken for a while because of all the changes that have been made on the Network layer.

To address this, the Network playground has been fixed to work with the latest changes. Additionally, it was updated to perform requests to GitHub v3 API, which allows users to play with a real API.

### DIFF
--- a/Alicerce.playground/Pages/Network.xcplaygroundpage/Contents.swift
+++ b/Alicerce.playground/Pages/Network.xcplaygroundpage/Contents.swift
@@ -1,21 +1,23 @@
 //: [Previous](@previous)
 
 import Alicerce
+import Result
 import PlaygroundSupport
 
 PlaygroundPage.current.needsIndefiniteExecution = true
 
 // Network Stack
 
-let network = Network.URLSessionNetworkStack()
+let network = Network.URLSessionNetworkStack(retryQueue: DispatchQueue(label: "com.alicerce.network.retry-queue"))
 
 network.session = URLSession(configuration: .default,
                              delegate: network,
                              delegateQueue: nil)
 
-// API Errors
+// API Error
 
-enum APIError: Error, Decodable {
+enum GitHubAPIError: Error, Decodable {
+
     case generic(message: String)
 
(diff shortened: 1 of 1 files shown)


Alicerce.playground/Pages/Network.xcplaygroundpage/Contents.swift
```diff
@@ -31,57 +33,170 @@ enum APIError: Error, Decodable {
...
+    }
+}
+
+// NetworkStore (via NetworkStack)
+
+extension GitHubResource: DecodableResource & PersistableResource & NetworkStoreStrategyFetchResource {
+
+    // DecodableResource
+
+    var decode: DecodeClosure { return { try JSONDecoder().decode(T.self, from: $0) } }
```
filipe-lemos: This could be the default implementation for a `protocol JSONDecodableResource: DecodableResource ` where `Internal: Decodable` and `External == Data`

p4checo: I thought about that, but actually we can't because we can't assume the encoding is JSON whenever the `External` type is `Data` 😜

filipe-lemos: Of course, that's why I suggested a new protocol to wrap that up. Something along the lines of:
```
protocol JSONDecodableResource: DecodableResource {}

extension JSONDecodableResource where Internal: Decodable, Extenal == Data {

    var decode: DecodeClosure { return { try JSONDecoder().decode(Internal.self, from: $0) } }
}
```

p4checo: Oops, missed that sorry 🙈 It would be more correct, indeed.

I'm personally not a big fan of instantiating a new decoder on each invocation, but we can easily work around this and even make it more generic by creating a `decoder: Decoder` property on the protocol. Then it would be assumed that the decoder you choose will be the right one to whatever format you are expecting.

PR #191, opened 2019-04-22
Multiple clean ups and minor additions 🧹

p4checo:
##  Changes: 

- Make `AnalyticsTracker` extend `AnyObject`

- Update `: class` to `: AnyObject` on protocols.

- Rename and reorganize some files.

- Remove leftover copyright headers on files.

- Add protocol extension on `EmptyExternalResource`, when `External == Data`.

- Handle 205 HTTP status codes on `URLSessionNetworkStack`, which also return an empty body on success.


Sources/Persistence/CoreData/CoreDataEntity.swift
```diff
@@ -1,6 +1,6 @@
 import CoreData
 
-public protocol CoreDataEntity: class, NSFetchRequestResult {
+public protocol CoreDataEntity: AnyObject & NSFetchRequestResult {
```
jwelton: Do we still need to specify `AnyObject` here? As for something to conform to `NSFetchRequestResult`, it must also conform to `NSObjectProtocol`, which is a class protocol.

Sources/PerformanceMetrics/Trackers/PerformanceMetrics+MultiTracker.swift
```diff
@@ -0,0 +1,142 @@
...
+    public class MultiTracker: PerformanceMetricsTracker {
+
+        /// The configured sub trackers.
+        public let trackers: [PerformanceMetricsTracker]
+
+        /// The tracker's tokenizer.
+        private let tokenizer = Tokenizer<Tag>()
+
+        /// The tracker's token dictionary, containing the mapping between internal and sub trackers' tokens.
+        private let tokens = Atomic<[Token<Tag> : [Token<Tag>]]>([:])
```
jwelton: Are we following the Mindera style guide for Alicerce? If so.... `= [:]` is preferred over `([:])` 💄

(sorry I know it was like this before too.... but just thought I would mention it while I saw it. Feel free to ignore if you want to 😄)

jwelton: As reference: https://github.com/Mindera/swift-style-guide

Sources/Persistence/CoreData/CoreDataEntity.swift
p4checo: You're absolutely right! Will remove redundant `AnyObject` conformance 😉

Sources/PerformanceMetrics/Trackers/PerformanceMetrics+MultiTracker.swift
p4checo: Yes, we try to follow our style guide, even though unfortunately it's not fully consistent yet throughout the project 😇 

Regarding this particular case, remember that `Atomic` is not a dictionary, but a wrapper over a dictionary (in this case). It's also not `ExpressibleByDictionaryLiteral`, so I don't think we can instantiate it like you mention 😆 (I just tried to add that conformance as an experiment, but failed miserably 😅)

jwelton: Ahh of course, yeah that makes sense 👍 

Thanks 😄

Sources/PerformanceMetrics/Trackers/PerformanceMetrics+MultiTracker.swift
```diff
@@ -0,0 +1,142 @@
...
+        /// Stops measuring the execution of a specific code block while attaching any additional metric metadata, by
+        /// propagating it to all registered sub trackers.
+        ///
+        /// - Parameters:
+        ///   - token: The metric's identifying token, returned by `start(with:)`.
+        ///   - metadata: The metric's metadata dictionary.
+        public func stop(with token: Token<Tag>, metadata: Metadata? = nil) {
+
+            guard let subTokens = tokens.modify({ $0.removeValue(forKey: token) }) else {
+                assertionFailure("🔥: missing sub tokens for token \(token)!")
```
renatorodrigues: Maybe, we can conform these log messages to the latest styleguide?

p4checo: Sure, will do 👍 

I also have to open a PR on our style guide one of these days 😅

PR #195, opened 2019-04-30
Create subspecs 🍕

### LINKED ISSUE #122: Create submodules / subspecs
Since framework users might only be interested in a subset of the framework's features, it makes sense to allow partial importing.

CocoaPods allows this to be done quite easily by using [Subspecs](https://guides.cocoapods.org/syntax/podspec.html#group_subspecs)

Carthage seems to have no official support for this (and it might even be [against their philosphy](https://github.com/Carthage/Carthage/issues/588#issuecomment-116137295))

Perhaps in the future it might make sense to break Alicerce into smaller µFrameworks, because this issue will tend to worsen as we add more functionalities to it 😅

p4checo:
At long last, here are the first subspecs of Alicerce! Sorry for the delay 😇

Essentially, the subspecs were a mapping of our already existing internal folder structure, with some exceptional groupings. Feel free to shout if you don't agree, or have another idea for the subspecs! 🙏 

Currently defined subspecs are:

- Core (Utils and Shared)

- Analytics

- Deeplinking

- Extensions

  + Foundation

  + QuartzCore

  + UIKit

- Logging

- Network

- Observers

- PerformanceMetrics

- Persistence

  + CoreData

- Resource

- Stores

- View

Closes #122.


p4checo:
> It seems weird have the extensions in the Core spec, as an ideia, we could have a subspec for the extensions, or perhaps breaking it more.

The Extensions in the Core spec are essentially `Foundation` extensions (e.g. `Data`, `Dictionary`, `DispatchQueue`, ...), and I thought that they could be considered as "Core" funcitonality.

Do you suggest we create an `Extensions` subspec? (or even `FoundationExtensions`, since we also have `UIKit` extensions in `UI` subspec, for instance)

PR #196, opened 2019-05-04
Remove iOS 9 support 👋

p4checo:
## Motivation

After some years supporting it, and with iOS 13 right around the corner, we felt it was time to drop support for iOS 9.

It also allows us to remove some legacy code (and tests), and finally be consistent in the iOS target version between main target and test targets.

## Changes

- Set deployment target to iOS 10 at project level, and inherit on all targets.

- Update `s.ios.deployment_target` on .podspec.

- Update Compatibility section in README.md.

- Remove all iOS 9 specific code.

- Revise code to use most up to date primitives.


Sources/Utils/Atomic.swift
```diff
@@ -32,6 +32,12 @@ public final class Atomic<Value> {
     public var value: Value {
         get { return withValue { $0 } }
         set(newValue) { swap(newValue) }
+        _modify {
```
renatorodrigues: Was this addition intended?

p4checo: Well, it was, even though it's not related to deprecating iOS 9. Technically it only requires Swift 5.0, but I forgot to put it in on the migration PR. 😇 

Do you prefer that I add it on a separate PR?

renatorodrigues: I'd suggest moving this to a separate PR. It's a _little big_ change to be introduced here.

p4checo: Fair enough, will do it 👍

PR #202, opened 2019-05-24
Clean up Network errors and type aliases

p4checo:
### Checklist

### Motivation and Context

The `Network.Value`, `Network.Error` and `Network.Configuration` weren’t particularly generic, as both were very specialized for working with a `URLSessionNetworkStack`.

In the same way, the `typealias`’es `Network.CompletionClosure` and `Network.AuthenticationCompletionClosure` could be placed "closer" to the API's that require them.

### Description

As such, a new `associatedtype Error` was added to the `NetworkStack` Protocol, and both `Network.Error` and `Network.Configuration` were nested inside `URLSessionNetworkStack`.

In the same way, the `Network.Value` was modified to also be generic over the `Response` value, so that it can be used on any other `NetworkStack`.

The `Network.CompletionClosure` and `Network.AuthenticationCompletionClosure` type aliases were also removed and new `typealias`’es were created in each protocol where they are useful (i.e. `NetworkStack`, `NetworkStore` and `AuthenticationChallengeHandler`).

The `Network.URLSessionNetworkStack`’s default `NetworkStore` Conformance was removed, so that it can be used with any other `NetworkStore` implementation (i.e. with any other `StoreError` type).

To increase flexibility and prepare for upcoming `Decodable` changes, the `NetworkPersistableStoreError` restrictions on each case’s error were removed, `.parse` was renamed to `.decode`, and `.other` was removed.

Additional handling was added to try and parse a custom `APIError` when the resource’s decode fails inside a `NetworkStore` fetch.

The `Network.URLSessionNetworkStack.Error`'s `.api` and `.http` were unified (again 😅), as `.api` is essentially a subset of `.http`, and the benefits of having them separated has proven to not be that much, especially if we're only interested in the status code (which is frequent).

Update `Network.URLSessionNetworkStack.Error.url` to use a more strongly typed `URLError`, which apparently is the "official" error type for `URLSession` errors, as
confirmed in: https://developer.apple.com/documentation/foundation/urlsession/datataskpublisher/failure


Sources/Network/URLSessionNetworkStack.swift
```diff
@@ -8,6 +8,38 @@ public extension Network {
...
         public typealias Response = URLResponse
 
+        public enum Error: Swift.Error {
+
+            public typealias APIError = Swift.Error
+            public typealias TotalRetriedDelay = Retry.Delay
+
+            case noRequest(Swift.Error)
+            case http(HTTP.StatusCode, URLResponse)
+            case api(APIError, HTTP.StatusCode, URLResponse)
```
p4checo: I'm a bit tempted to unify this case again with `.http`, as they're essentially the same (i.e. HTTP protocol error).

On our current project, most errors will be caught in `.api`, but it requires us to always parse an `APIError` from the payload.

However, it makes error handling a bit more cumbersome, because if we want to simply handle errors based on HTTP status code, we always have to check both `.http` and `.api`.

What do you guys think?

p4checo: Well, went ahead with the change, as nobody manifested an opinion on this 😢

Issue #203, opened 2019-08-20
Clear `viewModel` on `prepareForReuse` when using `ReusableViewModelView`


diogoeiras:
As the title implies, should we clear the `viewModel` on `prepareForReuse` when inheriting from `ReusableViewModelView`? With this, we would guarantee that the view/cell is on a pristine state when used again. It should be straightforward to implement as we only need to override `prepareForReuse` and set the `viewModel` to `nil`.

p4checo:
This is a cool idea! 👍 

However, I think that we should probably do this on our **base cells/reusable views**, and not on our **base protocols**. The reason is that the actual wiring (e.g. calling `setUpBindings` when the VM is set) is already done on actual implementations, and this behavior fits under the same category.

So, in practice we could have default `prepareForReuse` implementations that set the `viewModel = nil` on:
- `ViewModelCollectionViewCell`
- `ViewModelCollectionReusableView`
-  `TableViewCell`
- `ViewModelTableViewHeaderFooterView`

That way both the VM _set_ and _unset_ are in the same place ✨ 

Regarding your suggestion to change `ReusableViewModelView`, we _could_ add a `prepareForReuse` method to `ReusableViewModelView`, but it would only be useful for any custom views that are reusable and don't inherit from CollectionView/TableView cells/views. Then we could even create a protocol extension with a "default" `prepareForReuse` implementation that resets the VM, but we would still be missing the `setUpBindings` call on the VM's `didSet`, so it would be an "incomplete" Implementation. That being said, I think it's probably not worth it to change the `ReusableViewModelView` protocol, and simply perform the changes on the base cells/views.

Hope this made sense 😜

PR #204, opened 2019-09-13
Clear `viewModel` on base cells and reusable views

diogoeiras:
### Checklist

### Motivation and Context
To ensure the view/cell is on a pristine state when reused, the view model should be cleared on `prepareForReuse`. This default implementation also leads to the view model being set and unset in the same place. This PR closes issue [203](https://github.com/Mindera/Alicerce/issues/203).

### DIFF
--- a/Sources/View/ViewModelCollectionReusableView.swift
+++ b/Sources/View/ViewModelCollectionReusableView.swift
@@ -9,4 +9,11 @@ open class ViewModelCollectionReusableView<ViewModel>: CollectionReusableView, R
     }
 
     open func setUpBindings() {}
+
+    open override func prepareForReuse() {
+
+        viewModel = nil
+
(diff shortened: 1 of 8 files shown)


Sources/View/ViewModelCollectionReusableView.swift
```diff
@@ -9,4 +9,11 @@ open class ViewModelCollectionReusableView<ViewModel>: CollectionReusableView, R
     }
 
     open func setUpBindings() {}
+
+    open override func prepareForReuse() {
+
+        viewModel = nil
```
p4checo: why do you set this to `nil` **before** calling `super`? 🤔

diogoeiras: My mistake while copying and pasting stuff around. Should be in the inverse order.

PR #205, opened 2019-10-24
Improve SPM Integration, break into sub-modules 📦

portellaa:
### Checklist

### Motivation and Context
Since Swift Package Manager is becoming quite popular and Alicerce via CocoaPods supports subspecs, this is quite achievable with Swift Package Manager. It isn't very fancy, but at least is supported.

### Description
So each module as it's own target which will be referenced by each `library`. 

The name of the target it's only visible in the xcode project and the library is what is exposed to the products of libraries that includes Alicerce.
That's why i have named each `library` with the prefix `Alicerce` and gave the name of the module to the target.

Regarding the Extensions, since SPM doesn't supports the same as CocoaPods supports - sub-subspecs - the only way to achieve that is using a `library` with the name `AlicerceExtensionsFoundation`, ...

It would become something like:

```swift
// swift-tools-version:5.1

import PackageDescription

let package = Package(
    name: "Alicerce",
    platforms: [
        .iOS(.v10)
    ],
    products: [
        .library(
            name: "Alicerce",
            targets: [
                "Analytics",
                "Core",
                "DeepLinking",
                "ExtensionsFoundation",
                "ExtensionsQuartzCore",
                "ExtensionsUIKit",
                "Logging",
                "Network",
                "Observers",
                "PerformanceMetrics",
                "Persistence",
                "Resource",
                "Stores",
                "View"
            ]
        ),
        .library(name: "AlicerceAnalytics", targets: ["Analytics"]),
        .library(name: "AlicerceCore", targets: ["Core"]),
        .library(name: "AlicerceDeepLinking", targets: ["DeepLinking"]),
        .library(name: "AlicerceExtensionsFoundation", targets: ["ExtensionsFoundation"]),
        .library(name: "AlicerceExtensionsQuartzCore", targets: ["ExtensionsQuartzCore"]),
        .library(name: "AlicerceExtensionsUIKit", targets: ["ExtensionsUIKit"]),
        .library(name: "AlicerceLogging", targets: ["Logging"]),
        .library(name: "AlicerceNetwork", targets: ["Network"]),
        .library(name: "AlicerceObservers", targets: ["Observers"]),
        .library(name: "AlicercePerformanceMetrics", targets: ["PerformanceMetrics"]),
        .library(name: "AlicercePersistence", targets: ["Persistence"]),
        .library(name: "AlicerceResource", targets: ["Resource"]),
        .library(name: "AlicerceStores", targets: ["Stores"]),
        .library(name: "AlicerceView", targets: ["View"]),
    ],
    targets: [
        .target(name: "Analytics", dependencies: ["Core"]),
        .target(
            name: "Core",
            dependencies: ["ExtensionsFoundation", "ExtensionsQuartzCore", "ExtensionsUIKit"],
            path: "Sources",
            sources: ["Shared", "Utils"]
        ),
        .target(name: "DeepLinking", dependencies: ["Core"]),
        .target(name: "ExtensionsFoundation", path: "Sources/Extensions/Foundation"),
        .target(name: "ExtensionsQuartzCore", path: "Sources/Extensions/QuartzCore"),
        .target(name: "ExtensionsUIKit", path: "Sources/Extensions/UIKit"),
        .target(name: "Logging", dependencies: ["Core"]),
        .target(name: "Network", dependencies: ["Resource"]),
        .target(name: "Observers", dependencies: ["Core"]),
        .target(name: "PerformanceMetrics", dependencies: ["Core"]),
        .target(name: "Persistence", dependencies: ["Core", "Logging", "PerformanceMetrics"]),
        .target(name: "Resource", dependencies: ["Core"]),
        .target(
            name: "Stores",
            dependencies: ["Core", "Logging", "Network", "PerformanceMetrics", "Persistence", "Resource"]
        ),
        .target(name: "View", dependencies: ["Core"])
    ],
    swiftLanguageVersions: [ .version("5") ]
)
``` 

⚠️  ***important*** ⚠️ : 
Regarding the tests Target has `Objective-C` code, this is not supported by Swift Package Manager, as far as i know, i'm still investigating.
One thing i can do is exclude the CoreData path, since this is the one that includes `Objective-C` code.

**EDIT:** We are not adding support for Unit Tests via SPM for now, and will leave that as future work. CocoaPods is under the same situation.

All the suggestions are welcome, let's support modules on SPM 🚀 

#### Additions

- Remove `QuartzCore` extensions sub-sub-spec and merge them into `UIKit` extensions
- Add reference to Alicerce SPM package in workspace for easier iteration
- Add schemas for each library product (main + sub-libraries) that are now available via SPM
- Add stage in Travis CI to build the products via SPM


Package.swift
```diff
@@ -1,5 +1,58 @@
...
+                "DeepLinking",
+                "Extensions",
+                "Logging",
+                "Network",
+                "Observers",
+                "PerformanceMetrics",
+                "Persistence",
+                "Resource",
+                "Stores",
+                "View"]
```
p4checo: the closing `]` should be on the next line with the same indent level as `targets: [` 💅

Package.swift
```diff
@@ -1,5 +1,58 @@
...
+        .target(name: "PerformanceMetrics", dependencies: ["Core"]),
+        .target(name: "Persistence", dependencies: ["Core", "Logging", "PerformanceMetrics"]),
+        .target(name: "Resource", dependencies: ["Core"]),
+        .target(
+            name: "Stores",
+            dependencies: ["Core", "Logging", "Network", "PerformanceMetrics", "Persistence", "Resource"]
+        ),
+        .target(name: "View", dependencies: ["Core"])
+    ],
+    swiftLanguageVersions: [ .version("5") ]
```
p4checo: this should now be "5.1" 😉

p4checo:
Regarding your questions, @portellaa:

1. **Extensions sub-sub-specs equivalent in SPM**
Do you think it's worth it? IMO perhaps we can drop them for SPM 🤔 

2. **Test targets in SPM**
We currently don't publish our tests via CocoaPods either (although I have that almost done in a stash 😅), so we don't have to in SPM either, I guess 😇. I agree that would be nice, but perhaps we can leave that as a future improvement?  
The Obj-C limitation you mention only affects test targets?

Additionally, do you think we should add a stage on our pipeline to validate the SPM package description?

Package.swift
p4checo: apparently this only supports integer versions 🤷

Sources/Analytics/Trackers/Analytics+MultiTracker.swift
```diff
@@ -1,5 +1,9 @@
 import Foundation
 
+#if canImport(AlicerceCore)
```
filipe-lemos, line 3: I'm not sure I understand why we're wrapping the imports but not the code that depends on the imports, won't that fail to build? And if so, why not just let it fail on the `import` itself?

p4checo: These conditional imports are only used/relevant when we're building via SPM, where these modules actually exist and are necessary given the new sub-module dependencies specified in `Package.swift`. 

Otherwise, (e.g. via CocoaPods) these imports would break the build if not conditionally made (as the sub-modules wouldn't exist).

Also, considering our CI passed when building via both CocoaPods and SPM (which builds all modules), I'm not fully understanding your concern of the build failing... 🤔 Could you please be more specific? Or did I answer your question? 😄

.travis.yml
```diff
@@ -28,6 +28,11 @@ jobs:
         - xcodebuild $XCODE_ACTION -workspace "$XCODE_WORKSPACE" -scheme "$XCODE_SCHEME" -sdk "$XCODE_SDK" -destination "$XCODE_DESTINATION" -derivedDataPath "$XCODE_DERIVED_DATA_PATH" | xcpretty
       after_success:
         - bash <(curl -s https://codecov.io/bash) -D $XCODE_DERIVED_DATA_PATH -J "^$XCODE_SCHEME$"
+    - stage: SwiftPM Build
+      env:
+        - PLATFORM_TARGET=x86_64-apple-ios13.0-simulator
```
diogoeiras, line 33: I'm assuming that we're using 13 due to `osx_image` value. Couldn't we bump it up or would it be better to do on another pr?

diogoeiras (approved):
👌

.travis.yml
p4checo: You're assuming correctly 😉 And yes, I think that we should do that bump (Swift version, Xcode) in a separate PR

Package.swift
```diff
@@ -1,5 +1,75 @@
...
+                "AlicerceCore",
+                "AlicerceLogging",
+                "AlicerceNetwork",
+                "AlicercePerformanceMetrics",
+                "AlicercePersistence"
+            ],
+            path: "Sources/StackOrchestrator"
+        ),
+        .target(name: "AlicerceView", path: "Sources/View")
+    ],
```
portellaa, line 73: why not add the `.testTarget`?

portellaa: forget, already saw it 😂

Sources/Analytics/Trackers/Analytics+MultiTracker.swift
filipe-lemos: Ah, I had misunderstood the reason for these imports. Thank you for the explanation 🙇

filipe-lemos (approved):

PR #206, opened 2019-12-06
Add sizers for collection reusable views 📏

filipe-lemos:
### Checklist

### Motivation and Context
When laying out collection view cells (and other reusable views) it might be necessary to know the size of a cell before it is created.

### Description
The sizer is used to calculate the size of a view before it is created and added to the view hierarchy. This is useful in the case of collection view cells when the layout needs to know the size for a cell given its view model. It works by wrapping `systemLayoutSizeFitting` on a dummy view to which the view model is temporarily bound.

This also includes a cache to avoid repeating the same calculations unnecessarily when a previous value is known to be the same. It uses a dictionary of size values with a key generated by the view (conforming to `SizerViewModelView`) for a specific view model.


Sources/View/CollectionReusableViewSizer.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+final class CollectionReusableViewSizerCacheGroup {
+
+    private var sizers = [String: AnyObject]()
+
+    func sizeFor<View: UICollectionReusableView & SizerViewModelView>(
+        _ type: View.Type,
+        viewModel: CollectionReusableViewSizer<View>.ViewModel,
```
p4checo: perhaps we can define a `typealias Sizer = CollectionReusableViewSizer<View>` like above to reduce some of this verbosity 💅

Sources/View/CollectionReusableViewSizer.swift
```diff
@@ -0,0 +1,141 @@
...
+    associatedtype SizerCacheKey: Hashable = ConstantSizerCacheKey
+
+    static func sizerCacheKeyFor(viewModel: ViewModel) -> SizerCacheKey
+}
+
+extension SizerViewModelView where SizerCacheKey == ConstantSizerCacheKey {
+
+    static func sizerCacheKeyFor(viewModel: ViewModel) -> SizerCacheKey {
+
+        return SizerCacheKey()
```
p4checo: maybe we can inline this 💄

Sources/View/CollectionReusableViewSizer.swift
```diff
@@ -0,0 +1,141 @@
...
+        width: Dimension = .compressed,
+        height: Dimension = .compressed,
+        attributes: UICollectionViewLayoutAttributes? = nil
+    ) -> CGSize {
+
+        assert(view.viewModel == nil, "🧟‍♂️ Zombie view model detected! Kill it with fire!")
+
+        view.viewModel = viewModel // the sizer populates the dummy view with the view model
+
+        attributes.then(f: { view.apply($0) })
```
p4checo: perhaps we can use a simple `attributes.flatMap(view.apply)`? 💅

p4checo:
There are still some untested areas, it would be great if you could add some tests covering them 🙇

Sources/View/CollectionReusableViewSizer.swift
filipe-lemos: Above the generic was on the type while here it's on the func so the typealias would also need to be generic:

```
    typealias Sizer<View: UICollectionReusableView & SizerViewModelView> = CollectionReusableViewSizer<View>

    func sizeFor<View: UICollectionReusableView & SizerViewModelView>(
        _ type: View.Type,
        viewModel: Sizer<View>.ViewModel,
        width: Sizer<View>.Dimension = .compressed,
        height: Sizer<View>.Dimension = .compressed,
        attributes: UICollectionViewLayoutAttributes? = nil
    ) -> CGSize {
```

Not sure we gain much with this, do you think it's still worth it?

p4checo: Oh, I see. Missed that 🙈 never mind then!

PR #207, opened 2019-12-15
Implement backtracking on Router match ⏪

### LINKED ISSUE #167: Backtracking on `Route.Tree`'s `match`
Currently, the `Route.Tree`'s `match` has a limitation where it doesn't fallback (backtrack) to previous possibly valid paths (and thus handlers) when the chosen "matching path" is not completely matched.

Example:

Registered routes:

1. `scheme://host/path/one`
2. `scheme://host/*`

Received URL:

- `scheme://host/path/two`

In this above scenario, the current implementation will fail to match the URL, because when validating the component `path` it will give priority to the exact match on route no. 1, and when it fails on `two` (expects `one`) it will never backtrack to route no. 2 which has a wildcard on the first path component (and should handle the route).

Fixing this limitation will make the Router more robust and allow creating fallbacks and catch-all routes for some scenarios where a specialized handler might not yet be registered.

p4checo:
### Checklist

## Motivation and Description

Until now, our Router backing data structure (`Route.Tree`) wouldn't backtrack when it followed a tree path that ended up not matching, even if alternative matching routes were registered on the tree (although had less priority when being evaluated).

This was obviously a significant limitation that prevented users from creating more complex rule sets with multiple levels of valid fallbacks and ultimately catch all rules.

To address this, the existing tree data structure was completely redesigned to be a simpler and "purer" trie, with a single node type instead of nodes and edges. Being a single recursive type made all operations simpler, which included backtracking. Furthermore, it leaves more room to evolve in the future if we so desire, e.g. to add new types of route components, like regexes or pattern matched parameters.

Closes #167.

## Description

- Revise new `Route.Component`:

  + Rename `variable` to `parameter`, e.g. `:foo`.

  + Add new `wildcard` component, i.e. `*`.

  + Add new `catchAll` component, which can define or not a parameter, (e.g. `**`, `**foo`)

- Create new `Route.TrieNode`, replacing `Route.Tree` with some improvements:

  + Single node data structure, easier to recurse and scale.

  + Add new checks for invalid routes (i.,e. duplicate parameter names in routes, misplaced or conflicting catchAll's).

  + Made it conditionally conform to `Equatable` if the `Handler` is `Equatable`.

  + Add new "pretty print" `description` which displays the tree structure as a nice and easy do understand tree diagram.

- Rename `TreeRouter` to `Route.TrieRouter`, with some improvements:

  + Make it be backed by a single `Route.TrieNode` instead of one per
scheme mapped via a dictionary.

  + Make empty schemes (`nil` or `""`) be converted to wildcards (`*`),
to allow matching any scheme and define scheme level catch alls. Hosts
also preserve this behavior,

- Remake all related unit tests to make them more terse, easier to
maintain and scale.

- Add documentation.


Sources/DeepLinking/Route+TrieRouter.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        /// Routes the given route by matching it against the current trie, and optionally notifies routing success
+        /// with a custom payload from the route handler.
+        ///
+        /// - Parameters:
+        ///   - route: The route to route.
+        ///   - handleCompletion: The closure to notify routing success with custom payload from the route handler.
+        /// - Throws: A `TrieRouterError` error if the route wasn't found.
+        public func route(_ route: URL, handleCompletion: ((T) -> Void)? = nil) throws {
```
IvoPintodaSilva: Is there any benefit on calling this method without specifying the `handleCompletion`?

Tests/AlicerceTests/DeepLinking/Route+TrieNode_IsEmptyAndDescriptionTests.swift
```diff
@@ -0,0 +1,118 @@
...
+            │        │  └──┬ *
+            │        │     └──● handlerL
+            │        │
+            │        └──┬ **
+            │           └──● handlerM
+            │
+            ├──┬ **catchAll
+            │  └──● handlerN
+            │
+            └──● handlerO
```
IvoPintodaSilva: ✨

Sources/DeepLinking/Route+TrieRouter.swift
p4checo: This `handleCompletion` is useful if you want use the router handlers to: 
- perform some asynchronous work (e.g. read from a file) and be notified of its completion.
- return some payload (e.g. some blob, a type safe representation of the route)
- both of the above 😄 

The need for this closure based API is essentially because `route` itself is synchronous, which means the handlers are given the routes to handle synchronously as well.

However, there might be use cases where you are not interested in any payload or completion signaling, and simply want to know if the route was handled or not when calling `route` (which you know immediately). And that's essentially why it's defined as an optional in the API.

Does this make it clearer?

IvoPintodaSilva: Yep, thanks for the explanation 😃

Sources/DeepLinking/Route+Component.swift
```diff
@@ -2,70 +2,68 @@ import Foundation
...
 
             switch component.first {
             case ":"?:
-                let index = component.index(component.startIndex, offsetBy: 1)
-                let parameterName = String(component[index...])
+                let parameterIndex = component.index(component.startIndex, offsetBy: 1)
+                let parameterName = String(component[parameterIndex...])
 
-                assert(parameterName.isEmpty == false, "🔥 Path component's parameter name is empty!")
+                precondition(parameterName.isEmpty == false, "🔥 Path component's parameter name is empty!")
```
renatorodrigues: Should we really crash the program on -O builds? 🤔

Sources/DeepLinking/Route+TrieNode.swift
```diff
@@ -0,0 +1,403 @@
...
+
+        /// The node's catchAll child node.
+        private(set) var catchAll: CatchAll?
+
+        /// The node's handler.
+        private(set) var handler: Handler?
+
+        /// A Boolean value that indicates whether the node is empty.
+        public var isEmpty: Bool {
+            return constants.isEmpty && parameter == nil && wildcard == nil && catchAll == nil && handler == nil
```
renatorodrigues: Unnecessary return. ✨

Sources/DeepLinking/Route+TrieNode.swift
```diff
@@ -0,0 +1,403 @@
...
+            let handler: Handler
+        }
+
+        /// The node's constant child nodes.
+        private(set) var constants: [String: TrieNode]
+
+        /// The node's parameter child node.
+        private(set) var parameter: Parameter?
+
+        /// The node's wildcard chil node.
```
renatorodrigues: Typo on `chil`.

Sources/DeepLinking/Route+TrieNode.swift
```diff
@@ -0,0 +1,403 @@
...
+            && lhs.catchAll == rhs.catchAll
+            && lhs.handler == rhs.handler
+    }
+}
+
+// MARK: - CustomStringConvertible
+
+extension Route.TrieNode: CustomStringConvertible {
+
+    public var description: String {
```
renatorodrigues: Beautiful description. 👏

Sources/DeepLinking/Route+TrieRouter.swift
```diff
@@ -0,0 +1,228 @@
...
+    ///
+    /// Routes are registered with an associated handler, which on match handles the event and optionally invokes a
+    /// completion closure with an abitrary payload of type `T`. This allows handlers to perform asynchronous work even
+    /// though route matching is made synchronously.
+    ///
+    /// - Remark: Access to the backing trie tree data structure *is* synchronized, so all operations can safely be
+    /// called from different threads.
+    ///
+    /// - Note: https://en.wikipedia.org/wiki/Trie for more information.
+    public final class TrieRouter<T>: Router {
```
renatorodrigues: Great work. 👏

Sources/DeepLinking/Route+TrieRouter.swift
```diff
@@ -0,0 +1,228 @@
...
+                }
+            }
+        }
+
+        /// Unregisters the given route from the router, and returns the associated handler if found.
+        ///
+        /// - Parameter route: The route to unregister
+        /// - Throws: A `TrieRouterError` error if the route is invalid or wasn't found.
+        @discardableResult
+        public func unregister(_ route: URL) throws -> AnyRouteHandler<T> {
```
renatorodrigues: Missing `- Returns:` ;P

Tests/AlicerceTests/DeepLinking/Route+TrieNode_TestUtils.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+@testable import Alicerce
+
+extension Route.TrieNode {
+
+    static var empty: Route.TrieNode<Handler> { return .init() }
+
+    static func constant(_ name: String, node: Route.TrieNode<Handler>) -> Route.TrieNode<Handler> {
+
+        return .init(constants: [name: node])
```
renatorodrigues: You can get rid of these returns. ✨

Sources/DeepLinking/Route+Component.swift
p4checo: Hmm, that's a fair point... My idea was that these sort of invalid (route) components should be caught during development/testing and fail hard, however they might indeed slip into optimized builds and crash the app on scenarios where recovery is actually possible (e.g. by not allowing the route to be created on the router).

I think a less harsh approach would be to create a new error type to be thrown by the `Route.Component` initializer, which can then be mapped to an `invalidRoute` error on the router. What do you think?

renatorodrigues: I have mixed feelings on the approach here, because this represent a programming error, and as you said, this should be caught during development. With this in mind, I would go for an assertion.

On the other hand, one might want to create routes dynamically on runtime, I suppose. If so, your idea for a new error type is better because allows one to handle the failure.

I'm fine with both approaches, but after thinking about the subject while writing this comment, I'm more inclined for the new error type you suggested.

Sources/DeepLinking/Route+TrieRouter.swift
portellaa: 😍

Tests/AlicerceTests/DeepLinking/Route+TrieNode_IsEmptyAndDescriptionTests.swift
```diff
@@ -0,0 +1,118 @@
...
+            try node.add(["**catchAll"], handler: "handlerN")
+
+            try node.add([], handler: "handlerO")
+        } catch {
+            return XCTFail("🔥 Failed to add routes with error: \(error)!")
+        }
+
+        XCTAssertEqual(
+            node.description,
+            """
```
portellaa: 😍

Tests/AlicerceTests/DeepLinking/Route+TrieNode_MatchTests.swift
```diff
@@ -0,0 +1,366 @@
...
+            matchRoute: ["🏗"],
+            expectedHandler: "🎉"
+        )
+    }
+
+    // MARK: backtracking
+
+    func testMatch_WithRoutePartiallyMatchingMorePrioritaryRoute_ShouldBacktrackAndReturnNextMostPrioritaryHandler() {
+
+        // match: 🏗/🚧
```
portellaa: you are killing me with those representations 😂 😍

Sources/DeepLinking/Route+Component.swift
p4checo: Yeah, my train of thought was essentially the same.

I've made an experiment with a new custom error, but it then "breaks" `ExpressibleByStringLiteral` on `Route.Component` 😭. However, since it's mostly useful on the unit tests, we can simply move it to that target and use a `try!` underneath 😇

PR #215, opened 2020-03-18
Improve constrain API by defining ConstraintGroup initial isActive state

ricardomsm:
### Checklist

### Motivation and Context
The current constrain API didn't give us enough flexibility to work with various constraint groups for a view, as all constraints were activated after being set, which if we want something from some model layer to define which group is active, you'd to go though activation and deactivation which ampers performance.

### Description

- Simplified constrain API by removing constrain params from install and uninstall methods, as the ones being passed were those of local scope

- Added is active flag on constrain functions to enable if a group is active or not

### DIFF
--- a/Sources/AutoLayout/Constrain.swift
+++ b/Sources/AutoLayout/Constrain.swift
@@ -17,21 +17,20 @@ public final class ConstraintGroup {
     public init() { }
 
     fileprivate var constraints: [NSLayoutConstraint] = [] {
-        willSet { uninstall(constraints) }
-        didSet { install(constraints) }
+        willSet { uninstall() }
     }
 
     public var isActive: Bool {
         get { constraints.allSatisfy { $0.isActive } }
-        set { newValue ? install(constraints) : uninstall(constraints) }
+        set { newValue ? install() : uninstall() }
     }
 
-    private func install(_ constraints: [NSLayoutConstraint]) {
+    private func install() {
 
         NSLayoutConstraint.activate(constraints)
     }
 
-    private func uninstall(_ constraints: [NSLayoutConstraint]) {
+    private func uninstall() {
 
         NSLayoutConstraint.deactivate(constraints)
     }
@@ -41,6 +40,7 @@ public final class ConstraintGroup {
 public func constrain<A: LayoutItem>(
     _ a: A,
     replacing group: ConstraintGroup = .init(),
+    isActive: Bool = true,
     constraints: (A.ProxyType) -> Void
 ) -> ConstraintGroup {
 
@@ -51,6 +51,7 @@ public func constrain<A: LayoutItem>(
     constraints(a)
 
     group.constraints = context.constraints
+    group.isActive = isActive
 
     return group
 }
@@ -60,6 +61,7 @@ public func constrain<A: LayoutItem, B: LayoutItem>(
     _ a: A,
     _ b: B,
     replacing group: ConstraintGroup = .init(),
+    isActive: Bool = true,
     constraints: (A.ProxyType, B.ProxyType) -> Void
 ) -> ConstraintGroup {
 
@@ -71,6 +73,7 @@ public func constrain<A: LayoutItem, B: LayoutItem>(
     constraints(a, b)
 
     group.constraints = context.constraints
+    group.isActive = isActive
 
     return group
 }
@@ -81,6 +84,7 @@ public func constrain<A: LayoutItem, B: LayoutItem, C: LayoutItem>(
     _ b: B,
     _ c: C,
     replacing group: ConstraintGroup = .init(),
+    isActive: Bool = true,
     constraints: (A.ProxyType, B.ProxyType, C.ProxyType) -> Void
 ) -> ConstraintGroup {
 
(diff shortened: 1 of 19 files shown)


Sources/AutoLayout/Constrain.swift
```diff
@@ -17,21 +17,20 @@ public final class ConstraintGroup {
     public init() { }
 
     fileprivate var constraints: [NSLayoutConstraint] = [] {
-        willSet { uninstall(constraints) }
-        didSet { install(constraints) }
+        willSet { uninstall() }
```
p4checo: Should we look at `isActive` flag inside the `willSet`/`didSet` and `uninstall`/`install` accordingly? 🤔

Sources/AutoLayout/Constrain.swift
```diff
@@ -17,21 +17,20 @@ public final class ConstraintGroup {
...
     }
 
     public var isActive: Bool {
         get { constraints.allSatisfy { $0.isActive } }
-        set { newValue ? install(constraints) : uninstall(constraints) }
+        set { newValue ? install() : uninstall() }
     }
 
-    private func install(_ constraints: [NSLayoutConstraint]) {
+    private func install() {
```
p4checo: Since we're touching these, perhaps we could use the underlying UIKit naming and call these `activate`/`deactivate`? 💅

Sources/AutoLayout/Constrain.swift
```diff
@@ -41,6 +40,7 @@ public final class ConstraintGroup {
 public func constrain<A: LayoutItem>(
     _ a: A,
     replacing group: ConstraintGroup = .init(),
+    isActive: Bool = true,
```
p4checo: Maybe call this new parameter `activate`? (or `activated`) 💄 

My reasoning is that `isActive` refers to the group property, not the constrain API (which creates/updates a group)

Sources/AutoLayout/Constrain.swift
ricardomsm: Fine by me 🙂, since the naming will be even more consistent

Sources/AutoLayout/Constrain.swift
ricardomsm: Makes sense, I'll go with one of those

Sources/AutoLayout/Constrain.swift
ricardomsm: Hmm the way we have it now, I don't see that much again, unless Im missing something. I left the `uninstall` for when we're replacing a group to `uninstall` the previous from the view, and the `activate` flag will do the operation its supposed to (tbh I don't know the weight of deactivating twice if for example the flag is false). That would make sense if we change the ordering of the calls, but then the is active wouldn't `activate` or `deactivate` any more right?

p4checo: my point was most geared towards the following scenario where we set new constraints: 
1. it only makes sense to uninstall old constraints if `isActive == true` on `willSet`
2. it may make sense to install new constraints if `isActive == true` on `didSet`

Does it make sense?

ricardomsm: From that scenario yes, it makes sense, but with the `isActive` `set` property as is we'll be activating the same set of constraints twice won't we?

p4checo: yes, I think we have to keep the `isActive` behavior on `set`, and I also don't see any issue in activating/deactivating the constraints twice if we set both `constraints` and `isActive` at once. Do you? 🤔

ricardomsm: I couldn't find any problem 😅, and if it's ok with you it's ok with me, I thought there might some problem or performance amp with that double operation, sorry for being "chato" 😅

p4checo: I don't foresee any performance problem, as setting/unsetting the flag to the same value twice in quick succession should simply dirty the layout twice with the same value, which will be picked up in the next layout pass.

We should however ensure that when setting either one of the properties the correct behavior is preserved 👍 

And don't worry, you're not doing anything wrong! 😉 IMO these discussions are healthy and only lead to better solutions 💪

Sources/Persistence/PersistencePerformanceMetricsTracker.swift
```diff
@@ -105,7 +105,7 @@ public extension PersistencePerformanceMetricsTracker {
             try execute { result in
 
                 switch result {
-                case .success(let blobSize, let memSize):
+                case .success((let blobSize, let memSize)):
```
p4checo, line 108: I think you should exclude this file's changes from this PR, as it's Xcode 11.4 (Swift 5.2) only 😉

Tests/AlicerceTests/AutoLayout/CenterConstrainableProxyTestCase.swift
```diff
@@ -116,11 +116,48 @@ class CenterConstrainableProxyTestCase: BaseConstrainableProxyTestCase {
...
 }
 
 private extension CenterConstrainableProxyTestCase {
 
-    func expectedConstraints(view: UIView, to host: UIView) -> [NSLayoutConstraint] {
+    func expectedConstraints(
+        view: UIView,
+        to host: UIView,
+        active: Bool = true,
+        constant: CGFloat = .zero
```
p4checo, line 159: `constants` should be before `active` to match underlying API ordering 💅 
(more cases below)

PR #216, opened 2020-03-30
Add Locate property wrapper for ServiceLocator

beloso:
### Checklist
- [ ] I've built and run the project to see all new and existing tests pass

### Motivation and Context

With the new Property Wrappers in Swift 5.1 we can improve our `ServiceLocator` feature.
This PR proposes an approach for such a property wrapper.

### Description

I've added the new properly wrapper.
To test I am thinking about creating a Mock class with a property annotated with this wrapper.
Then I can register services in Test Methods and test that this Mock Class can be properly instantiated.

### DIFF
--- a/Sources/Utils/Locate.swift
+++ b/Sources/Utils/Locate.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+#if swift(>=5.1)
+@propertyWrapper
+public struct Locate<Service> {
+    private var service: Service
+    public init(name: String? = nil, locator: ServiceLocator = .shared) {
+        do {
+            self.service = try locator.get(name: name)
+        } catch {
+            fatalError(error.localizedDescription)
+        }
+    }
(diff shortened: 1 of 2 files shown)


Sources/Utils/Locate.swift
```diff
@@ -0,0 +1,23 @@
...
+            self.service = try locator.get(name: name)
+        } catch {
+            fatalError(error.localizedDescription)
+        }
+    }
+    public var wrappedValue: Service {
+        get { return service }
+        mutating set { service = newValue }
+    }
+    public var projectedValue: Locate<Service> {
```
dbeleza: I might be wrong 🙏, but I believe that considering that we are projecting _self_ and we do not have any public parameter or method, we could remove _projectedValue_ from here 🤔.
For more information: https://docs.swift.org/swift-book/LanguageGuide/Properties.html

PR #217, opened 2020-05-27
Remove require() function

mickael-domingues:
### Checklist

### Motivation and Context

- Remove unexpected program execution stoppage
- Better logging
- Better flow control

### Description

- Delete Optional.Swift and OptionalTests.Swift,
- Replace the use of .require() with guard statements,
- Add () on enum cases where the compiler would automatically tuple a  pattern but can not longer do so since Xcode 11.4.

More detail here: https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes
Thanks to @filipe-lemos for the hint.


Sources/Persistence/PersistencePerformanceMetricsTracker.swift
```diff
@@ -105,7 +105,7 @@ public extension PersistencePerformanceMetricsTracker {
             try execute { result in
 
                 switch result {
-                case .success(let blobSize, let memSize):
+                case .success((let blobSize, let memSize)):
```
p4checo, line 108: I think we can omit these for now because the CI hasn't been update do Xcode 11.4 yet, so this won't break our CI.

It's best to leave this to a proper Xcode 11.4 / Swift 5.2 migration 😉

mickael-domingues: Makes sense, I'll leave it as it was.

PR #218, opened 2020-06-01
Overhaul and deprotocolize Network 🛰

p4checo:
### Checklist

### Motivation and Context

As it was, our `NetworkStack` and it's main conforming type `URLSessionNetworkStack` were not straightforward to use, mostly because of the many different `Resource` protocols and respective
associated types. All these `Resource` protocols, while being broken into simple, focused and composable units, created lots of complexity due to the generic constraints they impose.

While initially designed for allowing diverse `NetworkStack`'s to be modeled, we have observed that most of the times users are simply interested in using a `URLSessionNetworkStack`.

The `NetworkStore` suffered from similar issues, by being tied to our huge `Resource` protocol family, while its main idea (and sole implementation) was a store that orchestrated both a network and persistence stacks.

Finally, `PersistenceStack` received a minor clean up and improvement to make it more flexible and powerful.

With all these things in mind, our `Resource` "family" was rethought to be composed of multiple witnesses, instead of multiple protocols. Witnesses provide the same abstraction of modeling behavior, without some of the limitations and complexities of protocols. By being generic
concrete types, they also allow creating default implementations via static properties and functions, making code reuse very simple.

The `NetworkStack` protocol was kept but greatly simplified and the `NetworkStore` protocol was discontinuated in favor of a `StackOrchestratorStore`, also simplified while making its purpose clearer.

Some concepts discussed on the amazing pointfree's videos were used, in particular protocol witnesses. Feel free to check this [talk](https://youtu.be/3BVkbWXcFS4) or their video series over at [pointfree.co](pointfree.co). 🤓🙏 

## Description

- Deprecate all `Resource` protocols, and replace most relevant ones with witnesses:

  + Create new `BaseRequestMaking` witness to replace `BaseRequestResource`.

  + Create new `ModelDecoding` witness to replace `DecodableResource`.

  + Create new `ErrorDecoding` witness to replace `ExternalErrorDecoderResource`.

- Deprecate `RequestAuthenticator` in favor of `URLRequestAuthenticator`, providing a simpler API.

- Deprecate `RequestInterceptor` in favour of a more powerful `URLSessionResourceInterceptor` defined at a Resource level, allowing much more flexibility and control. Add default conformances to `URLRequestAuthenticator`, `Network.URLSessionRetryPolicy`.

- Improve `Retry` infrastructure via new `Retry.State`, `Retry.Action.CompareClosure` and `Retry.Action.mostPrioritary` helper static function. Rename `Retry.Policy.retries` to `maxRetries`, `Retry.Policy.Backoff.Truncation.retries` to `maxRetries` and `Retry.Policy.Backoff.Truncation.delay` to `maxDelay`.

- Improve `HTTPResourceEndpoint`:

  + Convert `body` property into a throwable function, to allow handling encoding errors.

  + Convert `request` computed property into a throwable function, to propagate body encoding errors.

- Rename `Network.URLSessionNetworkStack.Error` to `Network.URLSessionError` and refine it by removing unnecessary `URLResponse` and using new `Retry.State`.

- Update `NetworkStack` to require just a `Resource` and `FetchError` associated types, and add protocol extension to perform decoding after a network fetch.

- Update Network playground with recent changes.

- Improve `PersistenceStack` to have a new `Key: Hashable` `associatedtype` instead of the hardcoded `Persistence.Key`, and renamed existing `Remote` to `Payload`.

- Deprecate `NetworkStore` in favor of `StackOrchestratorStore`, which requires a `NetworkStack` and `PersistenceStack associated types. Add a protocol extension to perform decoding after a fetch.

- Create new `StackOrchestrator` namespace.

- Rename `NetworkPersistableStore` to `StackOrchestrator.Store`, plus some internal adjustments.

- Rename `NetworkStorePerformanceMetricsTracker` to `StackOrchestratorPerformanceMetricsTracker`.

- Update `podspec`.

- Add some new rule exceptions on `.swiftlint.yml`


Sources/Network/HTTPResourceEndpoint.swift
```diff
@@ -59,8 +60,13 @@ public extension HTTPResourceEndpoint {
...
-        urlRequest.httpBody = body
+        urlRequest.httpBody = try makeBody()
 
         return urlRequest
     }
 }
+
+extension HTTPResourceEndpoint where Self: Encodable {
+
+    func makeBody() throws -> Data? { try JSONEncoder().encode(self) }
```
beloso, line 71: I am bit confused what's the intent here.
Should a model implement `HTTPResourceEndpoint`?

Sources/StackOrchestrator/StackOrchestrator+Store.swift
```diff
@@ -0,0 +1,278 @@
...
+        @discardableResult
+        public func fetch(
+            resource: Resource,
+            completion: @escaping FetchCompletionClosure<Payload>
+        ) -> Cancelable {
+
+            // TODO: change the callback structure to allow returning multiple values (+ completion)
+
+            switch resource.strategy {
+            case .networkThenPersistence:
```
beloso, line 51: Out of curiosity, and curiosity only, but can you give me a use case where persisted data is preferred over network data? Or am I misunderstanding the purpose of this?

beloso: I think I understood this. If network fails we fetch the persisted data.

beloso:
I like these changes. It reduces complexity on the network stack implementation.
Thanks for this.

Sources/Network/HTTPResourceEndpoint.swift
p4checo: In short, yes. You can have a model which is something like `PotatoRequest: Encodable`, which you then use to build your request's JSON body. 

This also works for more complex cases, like:

```swift
enum MyAPIEndpoint: HTTPResourceEndpoint {
    case .banana(BananaRequest)
    case .potato(PotatoRequest)
}

extension BananaRequest: Encodable { ... } 
extension PotatoRequest: Encodable { ... }

extension MyAPIEndpoint: Encodable {
 
    func encode(to encoder: Encoder) throws {
        switch self {
        } 
    }

```

Sources/StackOrchestrator/StackOrchestrator+Store.swift
p4checo: That's precisely it 😄

Sources/StackOrchestrator/StackOrchestrator+Store.swift
```diff
@@ -0,0 +1,278 @@
...
+            cancelable += networkFetch(
+                resource,
+                success: { [weak self ] value in
+
+                    self?.processNetworkValue(
+                        value,
+                        resource: resource,
+                        cancelable: cancelable,
+                        completion: completion
+                        )
```
IvoPintodaSilva, line 81: Misplaced `)` here and in line 100 💅

Sources/Network/Network+URLSessionResource.swift
```diff
@@ -0,0 +1,103 @@
...
+        public let baseRequestMaking: BaseRequestMaking<URLRequest>
+        public let errorDecoding: ErrorDecoding<Data, URLResponse>
+
+        public let interceptors: [URLSessionResourceInterceptor]
+
+        public internal(set) var retryState: Retry.State
+        public let retryActionPriority: Retry.Action.CompareClosure
+
+        public init(
+            baseRequestMaking: BaseRequestMaking<URLRequest>,
```
IvoPintodaSilva, line 16: Just to confirm my understanding on the use of protocol witnesses here:

Instead of having multiple `Resource` protocols, each for their purpose (data decoding, error decoding...), we now inject protocol witnesses.

These protocol witnesses are concrete generic types that make that desired functionality available.

By doing this we are able to remove generic constraint complexity from our protocols and concrete types.

Am I making sense or missing something? 😅

Sources/Network/HTTPResourceEndpoint.swift
p4checo: On a second thought, this extension is also assuming that if the endpoint is `Encodable` you want JSON encoding on your request's body... 😅 That being said, perhaps it's safer to not make it the "default" behavior. 🤔

Sources/Network/Network+BaseRequestMaking.swift
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+extension Network.BaseRequestMaking where Request == URLRequest {
+
+    /// Creates a new base request making witness from an `HTTPResourceEndpoint`.
+    /// - Parameter endpoint: the endpoint to make the base request from.
+    public static func endpoint(_ endpoint: HTTPResourceEndpoint) -> Self {
+
+        .init { handler in handler(.init { try endpoint.makeRequest() }) }
```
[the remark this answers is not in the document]
p4checo: That's a fair point. In my pending local changes, I've updated the inner `.init` (`Result`) to be explicit, so that it becomes easier to understand what's actually going on. 😉 

Would that be enough to make it clearer?

Sources/Network/Network+URLSessionResource.swift
p4checo: That's more or less it, yes 👍 

In essence, each protocol is converted to a witness that models/encapsulates the same behavior, which as you said greatly reduces generic requirements complexity when composed (vs protocol composition) without losing the actual abstraction. 

Additionally, as witnesses are concrete types, they give us some nice things "for free":

- you can have N witnesses for the same base type, whereas normally a type can only conform once to a protocol.

- you can create static helpers to instantiate default witness for a particular scenario (e.g. JSON decoding, or for your own Endpoint type).

- you can manipulate witnesses like any other type, because they are concrete types (e.g. add a `map`, `flatMap`, `contraMap`, ...)

Hope this helps

Sources/Network/Network+BaseRequestMaking.swift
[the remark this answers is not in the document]
IvoPintodaSilva: Yes, that makes it easier 👍 Thanks!

Sources/Network/HTTPResourceEndpoint.swift
p4checo: I've since removed this default implementation because it was making too much assumptions 😇

ricardomsm (approved):

IvoPintodaSilva (approved):

Documentation/Network.md
```diff
@@ -93,22 +152,44 @@ In the same way that network requests are commonly made via HTTP, it's also very
...
+
+The [`FetchAndDecodeError`][FetchAndDecodeError] is a simple error type used on `fetchAndDecode` calls and is used to differentiate between errors originating from either the fetch or decode operations. As such, it has just two cases which wrap an error each:
+* `fetch(Error`
+* `decode(Error)`
 
 ## Usage
 
 ### Setup
 
+Let's walk through a basic steps required to start making some requests with Alicerce. A similar setup is also available in a [Swift playground][Network.xcplaygroundpage] as a live example.
```
diogoeiras, line 191: Should be something like `Let's walk through the basic steps`

Sources/StackOrchestrator/StackOrchestrator+Store.swift
```diff
@@ -0,0 +1,284 @@
...
+        private func fetchNetworkFirst(
+            resource: Resource,
+            completion: @escaping FetchCompletionClosure<Payload>
+        ) -> Cancelable {
+
+            let cancelable = CancelableBag()
+
+            cancelable += networkFetch(
+                resource,
+                success: { [weak self ] value in
```
diogoeiras, line 74: Extra space after `self` 💄 (another case below)

Tests/AlicerceTests/Network/NetworkStackTestCase.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    private typealias NetworkStack = MockNetworkStack<Resource, Payload, Response, MockNetworkError>
+    private typealias NetworkDecoding<T> = ModelDecoding<T, Payload, Response>
+
+    private var stack: NetworkStack!
+
+    private var resource: Resource!
+    private let networkValue = Network.Value(value: "🌍", response: 1337)
+
+    override func setUpWithError() throws {
```
diogoeiras, line 21: Shouldn't we've a call to super? Same with `tearDownWithError`

diogoeiras (changes requested):

Documentation/Network.md
p4checo: Indeed! will fix

Sources/StackOrchestrator/StackOrchestrator+Store.swift
p4checo: 👀

Tests/AlicerceTests/Network/NetworkStackTestCase.swift
p4checo: For some reason the new autogenerated `XCTestCase` Unit Test templates don't have default `super` calls for `setUpWithError()` and `tearDownWithError()`, and as such I didn't implement them either 😅.

My guess is that it should no longer be necessary, because the "old" `setUp` and `tearDown` are still called under the hood, according to the documentation:

> `setUpWithError()`
>
> Before each test begins, XCTest calls setUpWithError(), followed by setUp(). If state preparation might throw errors, override setUpWithError(). XCTest marks the test failed when it catches errors, or skipped when it catches XCTSkip.

> `tearDownWithError()`
>
> After each test completes, XCTest calls tearDown(), followed by tearDownWithError(). If state cleanup might throw errors, override tearDownWithError(). XCTest marks the test failed when when it catches errors, or skipped when it catches XCTSkip.

So in essence I think Apple is telling us that we should use this to setup/tear down our tests and take advantage of the new throwing capabilities to mark tests as failed if necessary. Furthermore, I get the impression that the default (`super`) implementations for these two new methods don't do anything (otherwise Apple would add the `super` calls).

However, if you prefer I can add them as I think it won't hurt. It will add extra verbosity though. 😄

diogoeiras: Well, we don't have an hierarchy of more than one level (we're always subclassing `XCTestCase` directly) so it should be fine to not add them.

diogoeiras (approved):

filipe-lemos (approved):
👏 Great work! These changes should simplify the use of this lib. Left just some very minor comments

Documentation/Network.md
```diff
@@ -31,49 +32,107 @@ To be instantiated, a `URLSessionNetworkStack` requires the following dependenci
...
+The resource is the work unit of a Network stack, and abstracts an object that can have multiple representations (e.g. local, remote), capabilities and requirements. This allows each stack to require the set of behaviors it requires in a single concrete type, reducing the complexity of generic constraints.
+
+#### `URLSessionResource`
+
+To fetch data using a `URLSessionNetworkStack` we need to use a [`URLSessionResource`][URLSessionResource], which one can build by passing in the following dependencies:
+
+* A [`BaseRequestMaking<URLRequest>`](#baserequestmakingrequest) instance: attempts to generate a new *base* `URLRequest` asynchronously every time a request is scheduled for this Resource. This base request is then processed by the interceptor chain to possibly be enriched/modified before being scheduled on the network.
+
+* An [`ErrorDecoding<Data, URLResponse>`](#errordecodingpayload-metadata) instance: attempts to decode an arbitrary custom error (e.g. API error) from the payload and response if the response HTTP status code is not successful (2xx).
```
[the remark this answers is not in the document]
p4checo: well spotted, will reword

Documentation/Network.md
```diff
@@ -31,49 +32,107 @@ To be instantiated, a `URLSessionNetworkStack` requires the following dependenci
...
+The [`ErrorDecoding<Payload, Metadata>`][ErrorDecoding] type contains a single closure property `decode` which attempts to decode an arbitrary `Error` instance from a given `Payload?` and `Metadata` whenever a request completes with an **unsuccessful** HTTP status code (i.e. _not_ 2xx).
 
-  * [`ExternalResource`][ExternalResource] protocol: represents a `Resource` that also has an `External` representation (i.e. remote, like `Data`).
+The `Payload` is the main source of data to perform the decoding, but on some scenarios an additional `Metadata` can be helpful (e.g. information contained in response headers).
 
-    * [`RequestResource`][RequestResource] protocol: represents an `ExternalResource` that can be fetched by a `Request` object (i.e. like `URLRequest`).
+By being a struct and not a protocol (while modelling the same behavior), it greatly simplifies generics and allows easy default implementations via static factory methods. It currently provides an `.json()` helper to build an `ErrorDecoding<Data, _>` that attempts to decode a particular `Decodable` error type `E` encoded in JSON using a `JSONDecoder`.
```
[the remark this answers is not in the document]
p4checo: 🙈

Documentation/Network.md
```diff
@@ -93,22 +152,44 @@ In the same way that network requests are commonly made via HTTP, it's also very
...
 With the above protocols, we have the necessary infrastructure to fetch resources that require authentication. Additionally, since the authentication logic is only coupled to each Resource type and can (and should) be made asynchronously, it allows sharing the same network stack for all resources of an app, including authentication ones! 💪
 
-### Error
+### Decoding
+
+As mentioned above, the `NetworkStack` provides a `fetchAndDecode` function that automatically fetches and decodes a `Resource`, give  it's provided with a `ModelDecoding` witness. The failure type is a [`FetchAndDecodeError`](#fetchanddecodeerror) to allow the caller to differentiate the "origin" of the failure.
+
+#### `ModelDecoding<T, Payload, Metadata>`
+
+The [`ModelDecoding<T, Payload, Metadata>`][ModelDecoding] type contains a single closure property `decode` which attempts to decode an arbitrary `T` instance from a given `Payload` and `Metadata` whenever a request completes with an **successful** HTTP status code (i.e. 2xx, *except* 204 and 205 which expect empty bodies).
```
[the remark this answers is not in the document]
p4checo: 🙃

Sources/Network/Network+URLSessionResource.swift
```diff
@@ -0,0 +1,104 @@
...
+            errorDecoding: ErrorDecoding<Data, URLResponse>,
+            interceptors: [URLSessionResourceInterceptor],
```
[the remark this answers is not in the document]
p4checo: I don't see why not.

PR #222, opened 2020-10-21
Make url scheme and host match and register case insensitive

JOCuendet:
### Checklist

### Motivation and Context
The current implementation did not account for [RFC-3986](https://tools.ietf.org/html/rfc3986#section-6.2.2.1), and by doing so if the user entered an URL with a capital letters in the scheme or the host the match would fail.
By complying to the norm, this behaviour is now accounted for, and the issue is fixed.

### Description
Forcing the scheme and host from the user's input URL to be lowercased before parsing them to the match method.
Changed a UnitTest to accommodate to the base code changes.
Tested applying the code to a working project and running the app on a physical device to debug the deep links and universal links

### DIFF
--- a/Sources/DeepLinking/Route+TrieRouter.swift
+++ b/Sources/DeepLinking/Route+TrieRouter.swift
@@ -178,6 +178,7 @@ extension Route {
         private func parseAnnotatedRoute(_ route: URL) throws -> [Route.Component] {
 
             // use a wildcard for empty schemes/hosts, to match any scheme/host
+            // URL scheme and host comparison should be case insensitive in conformance to RFC-3986
             let schemeComponent = (route.scheme?.lowercased()).constantOrWildcardComponent
             let hostComponent = (route.host?.lowercased()).constantOrWildcardComponent
 
@@ -204,6 +205,7 @@ extension Route {
         private func parseMatchRoute(_ route: URL) throws -> MatchRoute {
 
             // use an empty string for empty scheme/host, to match wildcard scheme/host
+            // URL scheme and host comparison should be case insensitive in conformance to RFC-3986
             let schemeComponent = route.scheme?.lowercased() ?? ""
             let hostComponent = route.host?.lowercased() ?? ""
             let pathComponents = route.pathComponents.filter { $0 != "/" }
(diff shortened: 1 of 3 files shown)


filipe-lemos (approved):
Nice fix! 🎉 Perhaps add a couple of unit tests to `Route+TrieRouter_RouteTests` that check for case-insensitive matches.

ricardomsm (approved):

literallysofia (approved):

nbarreto6 (approved):

Sources/DeepLinking/Route+TrieRouter.swift
```diff
@@ -178,8 +178,8 @@ extension Route {
         private func parseAnnotatedRoute(_ route: URL) throws -> [Route.Component] {
 
             // use a wildcard for empty schemes/hosts, to match any scheme/host
```
p4checo, line 180: perhaps we could add a comment that host and scheme comparison should be case insensitive because of RFC3986 (same case below)

nqtavares (approved):

p4checo (changes requested):
Nice improvement! 💪 

Some suggestions:
- please add some new unit tests to check new behavior 🤓 
- the PR title (and final commit) could still be improved: perhaps "Make route scheme and host checks case insensitive"? 💅

Tests/AlicerceTests/DeepLinking/Route+TrieRouter_RouteTests.swift
```diff
@@ -130,6 +130,16 @@ class Route_TrieRouter_RouteTests: XCTestCase {
         XCTAssertRouteSucceeds(initial: [("/", testHandler)], route: "://")
     }
 
+    func testRoute_WithCaseInsensetiveScheme_shouldSucceed() {
```
p4checo, line 133: typo in "Insensetive", and "Should" should start with uppercase

also, perhaps:
`testRoute_WithCaseInsensitiveMatchingScheme_ShouldSucceed()`
`testRoute_WithCaseInsensitiveMatchingHost_ShouldSucceed()`

And maybe use some weird variations like `HttP` and `HoSt`, to make it clearer? (ie. not just CAPS vs caps)

p4checo (approved):
Nice one! Just some minor tweaks left IMO 👍

IvoPintodaSilva (approved):

PR #224, opened 2020-10-27
Add AutoLayout helper ConstraintGroupToggle

nbarreto6:
### Checklist

### Motivation and Context
ConstraintGroupToggle handles switching between ConstraintGroups that can't be active at the same time, deactivating the current active group and only after activating the new one.

Went back and forth on a few decisions, the biggest one is whether we should allow the closure that builds our ConstraintGroups to return an optional ConstraintGroup. 

` private let makeConstraintGroup: (T) -> ConstraintGroup?`

I ended up using this approach to avoid returning empty ConstraintGroups in cases we don't want to. 

With this in mind the other decision was wether nil would deactivate all constraints or keep the last active constraint active. 
I went with keeping the last constraint active, but not really sure if it's the right call.

If someone has feelings on these issues please let me know 😅

Kudos to @filipe-lemos for the backbone of the code

### Description
Add ConstraintGroupToggle to AutoLayout.Constraint
Add UnitTests

### DIFF
--- a/Sources/AutoLayout/Constrain.swift
+++ b/Sources/AutoLayout/Constrain.swift
@@ -41,6 +41,45 @@ public final class ConstraintGroup {
     }
 }
 
+public final class ConstraintGroupToggle<T: Hashable> {
+
+    private var constraintGroups: [T: ConstraintGroup] = [:]
+    private let makeConstraintGroup: (T) -> ConstraintGroup?
+
+    public init(initial: T? = nil, _ makeConstraintGroup: @escaping (T) -> ConstraintGroup?) {
+
+        self.makeConstraintGroup = makeConstraintGroup
+
+        if let initial = initial {
+            activate(initial)
+        }
+    }
+
+    public func activate(_ key: T) {
+
+        var previousActiveConstraint: ConstraintGroup?
+
+        constraintGroups.forEach {
+            
+            if $0 != key && $1.isActive {
+                $1.isActive = false
+                previousActiveConstraint = $1
+            }
+        }
+
+        if let constraintGroup = constraintGroups[key] {
+            constraintGroup.isActive = true
+        } else {
+            if let constraintGroup = makeConstraintGroup(key) {
+                constraintGroups[key] = constraintGroup
+                constraintGroup.isActive = true
+            } else {
+                previousActiveConstraint?.isActive = true
+            }
+        }
+    }
+}
+
 @discardableResult
 public func constrain<A: LayoutItem>(
     _ a: A,
--- a/Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
+++ b/Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
@@ -1,0 +1,254 @@
+import XCTest
+@testable import Alicerce
+
+
+class ConstraintGroupToggleTestCase: BaseConstrainableProxyTestCase {
+
+    private enum TestConstraintGroupKey: Hashable {
+        case first
+        case second
+        case third
+    }
+
+    func testNoInitialConstraint_WithEnumKey_ShouldNotActivateConstraints() {
+
+        var constraint1: NSLayoutConstraint!
+        var constraint2: NSLayoutConstraint!
+        var constraint3: NSLayoutConstraint!
+
+        let _ = ConstraintGroupToggle<TestConstraintGroupKey> { [weak self] in
+
+            guard let self = self else { return nil }
+
+            switch $0 {
+            case .first:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint1 = view0.bottom(to: host, offset: -100)
+                }
+
+            case .second:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint2 = view0.top(to: host, offset: -100)
+                }
+
+            case .third:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint3 = view0.leading(to: host, offset: -100)
+                }
+            }
+        }
+
+        XCTAssertNil(constraint1)
+        XCTAssertNil(constraint2)
+        XCTAssertNil(constraint3)
+    }
+
+    func testInitialConstraint_WithEnumKey_ShouldActivateFirstConstraint() {
+
+        var constraint1: NSLayoutConstraint!
+        var constraint2: NSLayoutConstraint!
+        var constraint3: NSLayoutConstraint!
+
+        let _ = ConstraintGroupToggle(initial: TestConstraintGroupKey.first) { [weak self] in
+
+            guard let self = self else { return nil }
+
+            switch $0 {
+            case .first:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint1 = view0.bottom(to: host, offset: -100)
+                }
+
+            case .second:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint2 = view0.top(to: host, offset: -100)
+                }
+
+            case .third:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint3 = view0.leading(to: host, offset: -100)
+                }
+            }
+        }
+
+        XCTAssert(constraint1.isActive)
+        XCTAssertNil(constraint2)
+        XCTAssertNil(constraint3)
+    }
+
+    func testActivateOneConstraint_WithEnumKey_ShouldDeactivateFirstConstraintAndActivateSecondConstraint() {
+
+        var constraint1: NSLayoutConstraint!
+        var constraint2: NSLayoutConstraint!
+        var constraint3: NSLayoutConstraint!
+
+        let group = ConstraintGroupToggle(initial: TestConstraintGroupKey.first) { [weak self] in
+
+            guard let self = self else { return nil }
+
+            switch $0 {
+            case .first:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint1 = view0.bottom(to: host, offset: -100)
+                }
+
+            case .second:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint2 = view0.top(to: host, offset: -100)
+                }
+
+            case .third:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint3 = view0.leading(to: host, offset: -100)
+                }
+            }
+        }
+
+        group.activate(.second)
+
+        XCTAssertFalse(constraint1.isActive)
+        XCTAssert(constraint2.isActive)
+        XCTAssertNil(constraint3)
+    }
+
+    func testActivateTwoConstraints_WithEnumKey_ShouldDeactivateFirstAndSecondConstraintAndActivateThirdConstraint() {
+
+        var constraint1: NSLayoutConstraint!
+        var constraint2: NSLayoutConstraint!
+        var constraint3: NSLayoutConstraint!
+
+        let group = ConstraintGroupToggle(initial: TestConstraintGroupKey.first) { [weak self] in
+
+            guard let self = self else { return nil }
+
+            switch $0 {
+            case .first:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint1 = view0.bottom(to: host, offset: -100)
+                }
+
+            case .second:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint2 = view0.top(to: host, offset: -100)
+                }
+
+            case .third:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint3 = view0.leading(to: host, offset: -100)
+                }
+            }
+        }
+
+        group.activate(.second)
+        group.activate(.third)
+
+        XCTAssertFalse(constraint1.isActive)
+        XCTAssertFalse(constraint2.isActive)
+        XCTAssert(constraint3.isActive)
+    }
+
+    func testReactivateConstraint_WithEnumKey_ShouldDeactivateSecondAndThirdConstraintAndActivateFirstConstraint() {
+
+        var constraint1: NSLayoutConstraint!
+        var constraint2: NSLayoutConstraint!
+        var constraint3: NSLayoutConstraint!
+
+        let group = ConstraintGroupToggle(initial: TestConstraintGroupKey.first) { [weak self]  in
+
+            guard let self = self else { return nil }
+
+            switch $0 {
+            case .first:
+                return constrain(self.host, self.view0) { host, view0 in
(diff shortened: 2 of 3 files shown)


literallysofia (approved):

Sources/AutoLayout/Constrain.swift
```diff
@@ -41,6 +41,45 @@ public final class ConstraintGroup {
...
+        if let initial = initial {
+            activate(initial)
+        }
+    }
+
+    public func activate(_ key: T) {
+
+        var previousActiveConstraint: ConstraintGroup?
+
+        constraintGroups.forEach {
```
IvoPintodaSilva, line 62: This could be `constraintGroups.filter { $0 != key && $1.isActive }`.

And since we're only supposed to have one active constraint group at a time, we would be able to assert the `.count` and make it a bit safer maybe?

Sources/AutoLayout/Constrain.swift
```diff
@@ -41,6 +41,45 @@ public final class ConstraintGroup {
...
+        }
+
+        if let constraintGroup = constraintGroups[key] {
+            constraintGroup.isActive = true
+        } else {
+            if let constraintGroup = makeConstraintGroup(key) {
+                constraintGroups[key] = constraintGroup
+                constraintGroup.isActive = true
+            } else {
+                previousActiveConstraint?.isActive = true
```
IvoPintodaSilva, line 77: It feels a bit like we're hiding an error 🤔 Should we propagate an error is a user tries to activate a `constraintGroup` with an invalid key?

IvoPintodaSilva (commented):
Nice work 👍 Just left a few comments regarding constraint activation errors.

Sources/AutoLayout/Constrain.swift
filipe-lemos, line 77: As a user of this API if for some reason the `makeConstraintGroup` returns `nil` I would expect the behaviour of `activate` to still disable all the previous constraints, no? It just wouldn't apply any constraints, same as if we return `ConstraintGroup()` and same as if it's the initial `key`.

Maybe `constraintGroups` should be `[T: ConstraintGroup?]` to avoid re-running `makeConstraintGroup` when it returns `nil`.

nqtavares (approved):

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,254 @@
...
+        XCTAssertNil(constraint3)
+    }
+
+    func testActivateOneConstraint_WithEnumKey_ShouldDeactivateFirstConstraintAndActivateSecondConstraint() {
+
+        var constraint1: NSLayoutConstraint!
+        var constraint2: NSLayoutConstraint!
+        var constraint3: NSLayoutConstraint!
+
+        let group = ConstraintGroupToggle(initial: TestConstraintGroupKey.first) { [weak self] in
```
filipe-lemos, line 85: Maybe name this `groupToggle` instead of just `group` as it's a toggle of constraint groups and not just one constraint group.

Also these constraints and toggle are the same in all or almost all the tests so maybe they can  be set up in the `XCTest.setUp` override.

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,254 @@
...
+
+            default:
+                return nil
+            }
+
+        }
+
+        group.activate("constraint3")
+
+        XCTAssert(constraint1.isActive)
```
filipe-lemos, line 251: As pointed out above, wouldn't we expect this group to be disabled?

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,254 @@
...
+                }
+
+            case .third:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint3 = view0.leading(to: host, offset: -100)
+                }
+            }
+        }
+
+        group.activate(.second)
```
filipe-lemos, line 142: Maybe assert the values after this `activate`?

Sources/AutoLayout/Constrain.swift
nbarreto6: Does look better :)

Sources/AutoLayout/Constrain.swift
nbarreto6: @IvoPintodaSilva @filipe-lemos I had tried both of these iterations:

Filipe's was my first iteration, I went with keeping the last constraint mainly because toggle made me feel that there should always be an active ConstraintGroup, that being said we can init without a ConstraintGroup active.   

In regards to Ivo's approach I also considered it and it might be the best way to lead the programmer away from code smells, like using Non Finite Key Sets.

I am leaning towards returning an error, but not sure if I should undo the previous constraints in that case.

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
nbarreto6: will do 👍

Sources/AutoLayout/Constrain.swift
IvoPintodaSilva: In my opinion, from a caller's perspective, if a method that I'm calling yields an error, I expect no changes in the internal state.

filipe-lemos: But that's assuming a value of `nil` in a `makeConstraintGroup` closure is an error, I would consider that as "for this key there are no constraints"

nbarreto6: I have now migrated a few use cases in our app and the only situation where I returned nil was because I have to capture self `guard let self = self else { return nil }`

I considered having the closure itself throw `makeConstraintGroup: (T) throws -> ConstraintGroup?`, but felt a bit like overkill.

So decision time, I think we can go with Filipe's more general approach and iterate in the future if we feel the need. If there are now further objections 😄

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,254 @@
...
+
+class ConstraintGroupToggleTestCase: BaseConstrainableProxyTestCase {
+
+    private enum TestConstraintGroupKey: Hashable {
+        case first
+        case second
+        case third
+    }
+
+    func testNoInitialConstraint_WithEnumKey_ShouldNotActivateConstraints() {
```
p4checo, line 13: format usually follows: `test<MethodOrPropertyName>_When<Conditions>_Should<ExpectedOutcome> {`

I think we're mixing the components a bit here 😅 
we're testing the `init`, and the conditions are e.g. "no initial constraint active". right?

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,254 @@
...
+                return nil
+            }
+
+        }
+
+        group.activate("constraint3")
+
+        XCTAssert(constraint1.isActive)
+        XCTAssertNil(constraint2)
+    }
```
p4checo, line 253: should we try to add a test case where the constraint groups are incompatible (thus need to be mutually exclusive), so that we can test that the mechanics are validated for a use case that resembles reality more?

p4checo (changes requested):
Nice work! 💪 
Got some questions regarding the `init` and overall setup API, and also think the tests can still be a bit improved IMO. 😉

Sources/AutoLayout/Constrain.swift
```diff
@@ -41,6 +41,45 @@ public final class ConstraintGroup {
     }
 }
 
+public final class ConstraintGroupToggle<T: Hashable> {
+
+    private var constraintGroups: [T: ConstraintGroup] = [:]
+    private let makeConstraintGroup: (T) -> ConstraintGroup?
+
+    public init(initial: T? = nil, _ makeConstraintGroup: @escaping (T) -> ConstraintGroup?) {
```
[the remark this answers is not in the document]
nbarreto6: I had one iteration where I had this:

```
public init(initial: T? = nil, _ constraintGroups: [T: ConstraintGroup]) {

        self.constraintGroups = constraintGroups

        if let initial = initial {
            activate(initial)
        } else {
            constraintGroups.forEach {
                $1.isActive = false
            }
        }
    }
```

With this approach if the ConstraintGroups are initialised without `active: false` all the constraints will be active for a bit before they are inactivated again. The lazy way I don't think we have that problem.

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,254 @@
...
+        var constraint3: NSLayoutConstraint!
+
+        let _ = ConstraintGroupToggle<TestConstraintGroupKey> { [weak self] in
+
+            guard let self = self else { return nil }
+
+            switch $0 {
+            case .first:
+                return constrain(self.host, self.view0) { host, view0 in
+                   constraint1 = view0.bottom(to: host, offset: -100)
```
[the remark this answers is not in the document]
nbarreto6: nope 👍

Sources/AutoLayout/Constrain.swift
```diff
@@ -41,6 +41,45 @@ public final class ConstraintGroup {
...
+
+        constraintGroups.forEach {
+            
+            if $0 != key && $1.isActive {
+                $1.isActive = false
+                previousActiveConstraint = $1
+            }
+        }
+
+        if let constraintGroup = constraintGroups[key] {
```
diogoeiras, line 70: Shouldn't we check if we've this key earlier on the func? Also, shouldn't the func itself return something (like a bool) so it's know if we activated it successfully? 
Edt: Forget it, I see we add it after if it doesn't exist. Maybe make shouldn't return an optional?

JOCuendet (approved):

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,193 @@
+import XCTest
+@testable import Alicerce
+
+
```
p4checo, line 4: too much empty space 🚓

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+        constraintGroupToggle.activate(.second)
+
+        XCTAssertFalse(constraint1.isActive)
+        XCTAssert(constraint2.isActive)
+        XCTAssertFalse(constraint3.isActive)
+
+        constraintGroupToggle.activate(.first)
+
+        XCTAssertTrue(constraint1.isActive)
```
p4checo, line 104: `XCTAssert`

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+
+private enum TestConstraintGroupKey: Hashable {
+    case first
+    case second
+    case third
+}
+
+private final class MockContraintGroup: ConstraintGroup {
+
+    private let activateClosure: () -> Void
```
p4checo, line 169: no need to be so concerned with encapsulation in mocks, and we don't need to pass them on `init`, as it makes them more cumbersome to use.

IMO this could be a single optional `var` closure like `var didSetIsActive: ((Bool) -> Void)?`, which we then invoke on the `isActive` property `didSet`. This allows users to use mocks easily and only intercept calls when necessary.

We also try to avoid that mocks have "business logic", like `isConstraintActive` and private functions.

Does it make sense?

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+            initial: "constraint1",
+            constraintGroups: ["constraint1": one, "constraint2": two]
+        )
+
+        nonFiniteConstraintGroupToggle.activate("constraint2")
+
+        wait(
+            for: [activateClosureOne, deactivateClosureOne, activateClosureTwo],
+            timeout: 1,
+            enforceOrder: true
```
p4checo, line 156: this is nice 👏

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
```diff
@@ -0,0 +1,193 @@
...
+    private var constraintGroup3: ConstraintGroup!
+
+    private var constraint1: NSLayoutConstraint!
+    private var constraint2: NSLayoutConstraint!
+    private var constraint3: NSLayoutConstraint!
+
+    override func setUp() {
+        super.setUp()
+
+        constraintGroup1 = constrain(self.view0, self.view1, activate: false) { view0, view1 in
```
p4checo, line 21: maybe we should test a scenario where we pass in the groups all activated, and expect it to deactivate them? 🤔

p4checo (approved):
Nice work! 💪 
just a couple of minor tweaks left IMO😉

IvoPintodaSilva (approved):

Sources/AutoLayout/Constrain.swift
```diff
@@ -41,6 +41,31 @@ public final class ConstraintGroup {
...
+    private var constraintGroups: [T: ConstraintGroup] = [:]
+
+    public init(initial: T? = nil, constraintGroups: [T: ConstraintGroup]) {
+
+        self.constraintGroups = constraintGroups
+
+        if let initial = initial {
+            activate(initial)
+        } else {
+            constraintGroups.forEach { $1.isActive = false }
```
filipe-lemos, line 55: Maybe we could have a `public func deactivate` that would turn off all constraint groups and be used here as well

filipe-lemos (approved):

diogoeiras (approved):

p4checo (approved):

Tests/AlicerceTests/AutoLayout/ConstraintGroupToggleTestCase.swift
p4checo: As discussed, `ConstraintGroup`'s internal mechanics make it harder to mock and test, and as such let's leave that problem for the future 😉

PR #227, opened 2021-04-04
Add `String` multi-replace via `Scanner` ⏩

p4checo:
### Checklist

### Motivation and Context

Sometimes we need to replace multiple characters in a given `String`, but both `replacingOccurrences()` and `replacingCharacters` operate on a single `String`/`Character`, requiring multiple passes and thus becoming very inefficient. One such example is to replace all line breaking characters in a string into their non line breaking version, which requires 6 substitution "passes" (space, hyphen, em dash, en dash, question mark and closing brace).

By using a `Scanner` as a matching mechanism, we can implement multi-replace in a single pass on the string, greatly improving efficiency.

### Description

- Add new `String.replacingOccurrencesOfCharacters(in:skippingCharactersIn:)` extension to allow replacing multiple characters in a string in a single pass.

- Add new `String.nonLineBreaking()` extension to convert a string into a non line breaking version.

### DIFF
--- a/Sources/Extensions/Foundation/String.swift
+++ b/Sources/Extensions/Foundation/String.swift
@@ -59,3 +59,91 @@ public extension String {
         dump(x, to: &self)
     }
 }
+
+extension String {
+
+    /// Replaces occurrences of multiple `Character`s with corresponding `String` values using the given mapping, while
+    /// skipping (filtering out) an optional set of characters from the output. Being backed by a `Scanner`, a single
+    /// pass is made over the receiver.
+    ///
+    /// - Parameters:
+    ///   - replacementMap: A dictionary containing the replacement mapping `Character` -> `String`.
+    ///   - charactersToBeSkipped: An optional set of characters to skip (i.e. filter out from the input).
+    /// - Returns: A modified version of the receiver with the replacement mapping applied.
+    public func replacingOccurrencesOfCharacters(
+        in replacementMap: [Character: String],
+        skippingCharactersIn charactersToBeSkipped: CharacterSet? = nil
+    ) -> String {
+
+        guard !replacementMap.isEmpty else { return self }
+
+        let matchSet = CharacterSet(charactersIn: replacementMap.keys.reduce(into: "") { $0 += String($1) })
+            .union(charactersToBeSkipped ?? CharacterSet())
+
+        var final = ""
+
+        let scanner = Scanner(string: self)
+        scanner.charactersToBeSkipped = charactersToBeSkipped
+
+        while !scanner.isAtEnd {
+
+            // copy everything until finding a character to be replaced or skipped
+            var collector: NSString? = ""
+            if scanner.scanUpToCharacters(from: matchSet, into: &collector), let collector = collector {
+                final.append(collector as String)
+            }
+
+            // exit early if we're already at the end
+            guard !scanner.isAtEnd else { break }
+
+            // skip characters if needed
+            if let charactersToBeSkipped = charactersToBeSkipped {
+                scanner.scanCharacters(from: charactersToBeSkipped, into: nil)
+            }
+
+            // find and replace matching character if needed
+            replacementMap
+                .first { match, _ in scanner.scanString(String(match), into: nil) }
+                .flatMap { _, replacement in final.append(replacement) }
+        }
+
+        return final
+    }
+}
+
+extension String {
+
+    public static let nonBreakingSpace = String(Character.nonBreakingSpace)
+    public static let nonBreakingHyphen = String(Character.nonBreakingHyphen)
+    public static let wordJoiner = String(Character.wordJoiner)
+    public static let emDash = String(Character.emDash)
+    public static let enDash = String(Character.enDash)
+
+    /// Returns a non line breaking version of `self`. Line breaking characters occurrences are replaced with
+    /// corresponding non line breaking variants when existent. Otherwise, word joiner characters are attached to them
+    /// to make them non line breaking.
+    ///
+    /// The character mapping is:
+    ///  - space (" ") -> non breaking space (`U+2028`)
+    ///  - hyphen ("-") -> non breaking hyphen (`U+00A0`)
+    ///  - em dash ("—") -> word joiner (`U+2060`) + em dash + word joiner (`U+2060`)
+    ///  - em dash ("–") -> word joiner (`U+2060`) + en dash + word joiner (`U+2060`)
+    ///  - question mark ("?") -> question mark + word joiner (`U+2060`)
+    ///  - closing brace ("}") -> closing brace + word joiner (`U+2060`)
+    ///
+    /// - Returns: A modified version of the receiver without line breaking characters.
+    public func nonLineBreaking() -> String {
+
+        replacingOccurrencesOfCharacters(
+            in: [
+                " ": String.nonBreakingSpace,
+                "-": String.nonBreakingHyphen,
+                .emDash: String([.wordJoiner, .emDash, .wordJoiner]),
+                .enDash: String([.wordJoiner, .enDash, .wordJoiner]),
+                "?": "?" + .wordJoiner,
+                "}": "}" + .wordJoiner,
+            ],
+            skippingCharactersIn: nil
+        )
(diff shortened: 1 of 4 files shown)


nbarreto6 (approved):

Sources/Extensions/Foundation/String.swift
```diff
@@ -59,3 +59,91 @@ public extension String {
...
+        replacingOccurrencesOfCharacters(
+            in: [
+                " ": String.nonBreakingSpace,
+                "-": String.nonBreakingHyphen,
+                .emDash: String([.wordJoiner, .emDash, .wordJoiner]),
+                .enDash: String([.wordJoiner, .enDash, .wordJoiner]),
+                "?": "?" + .wordJoiner,
+                "}": "}" + .wordJoiner,
+            ],
+            skippingCharactersIn: nil
```
filipe-lemos, line 146: I believe we want to skip `.newlines` here.

An alternative approach would be to replace occurences of `"\n"`, `"\r"`, [etc...](https://developer.apple.com/documentation/foundation/characterset/1780325-newlines) with empty strings and we could avoid the whole `skippingCharactersIn` which might not be super-intuitive for a general tool.

p4checo: I explicitly removed `.newlines` from being passed by default (from our private implementation, that is 🤭), precisely because it previously had weird behavior when newline characters were present in the middle of the string. Essentially, it preserved newlines (in this case) if they appeared **before** a match.

That led me to adding these fixes ([fix1](https://github.com/Mindera/Alicerce/pull/227/files/0f7a36ac4d757a72ccb61ce700bf27811c458e41#diff-b69213790c7e203256a581e51d5a09b68d06af47e7c3aa10c7845448a38b7b76R81), [fix2](https://github.com/Mindera/Alicerce/pull/227/files/0f7a36ac4d757a72ccb61ce700bf27811c458e41#diff-b69213790c7e203256a581e51d5a09b68d06af47e7c3aa10c7845448a38b7b76R99-R102)) which correctly ignores skipped characters if they appear in the middle of the text, and makes `skippingCharactersIn` behave as expected from an API perspective:

```swift
let matchSet = CharacterSet(charactersIn: replacementMap.keys.reduce(into: "") { $0 += String($1) })
    .union(charactersToBeSkipped ?? CharacterSet())
```

```swift
// skip characters if needed
if let charactersToBeSkipped = charactersToBeSkipped {
    scanner.scanCharacters(from: charactersToBeSkipped, into: nil)
}
```

However, after adding the fixes, then all newline characters were effectively skipped, but never replaced (we have no rule for that) - meaning words would get glued if on separate lines, for instance. Considering that the purpose of this helper is to avoid text from being cut when not _desired_, I think that stripping explicit newlines from the text is perhaps a different transformation. That's why in the end I decided that newlines would be preserved.

Furthermore, replacing newlines with a non breaking space (which seems like the best candidate), might not always produce the desired/expected output either, because in the same way that gluing words/elements in consecutive lines can be an artifact, so can "separating" them. Do you agree?

**PS**: After doing these tests once again, I noticed that [fix2](https://github.com/Mindera/Alicerce/pull/227/files/0f7a36ac4d757a72ccb61ce700bf27811c458e41#diff-b69213790c7e203256a581e51d5a09b68d06af47e7c3aa10c7845448a38b7b76R99-R102) it not needed, as adding the characters to be skipped to the `matchSet` is enough for them to be skipped (i.e. no need for the explicit `scanner.scanCharacters(from: charactersToBeSkipped, into: nil)` after a match is detected).

diogoeiras (approved):

Sources/Extensions/Foundation/String.swift
filipe-lemos: Fair enough 👌

Sources/Extensions/Foundation/String.swift
```diff
@@ -59,3 +59,88 @@ public extension String {
...
+    /// corresponding non line breaking variants when existent. Otherwise, word joiner characters are attached to them
+    /// to make them non line breaking.
+    ///
+    /// - Important: Any existing newline characters are preserved.
+    ///
+    /// The character mapping is:
+    ///  - space (" ") -> non breaking space (`U+2028`)
+    ///  - hyphen ("-") -> non breaking hyphen (`U+00A0`)
+    ///  - em dash ("—") -> word joiner (`U+2060`) + em dash + word joiner (`U+2060`)
+    ///  - em dash ("–") -> word joiner (`U+2060`) + en dash + word joiner (`U+2060`)
```
aclima93, line 127: 💅 i think this line is supposed to be "/// - en dash" instead of "em"?

Tests/AlicerceTests/Extensions/Foundation/StringTestCase.swift
```diff
@@ -66,5 +66,110 @@ class StringTestCase: XCTestCase {
...
+            \nThe\(String.nonBreakingSpace)quick\(String.nonBreakingHyphen)brown\u{85}\
+            \(String([.wordJoiner, .emDash, .wordJoiner]))fox\n\
+            \(String([.wordJoiner, .enDash, .wordJoiner]))jumps\
+            ?\(String.wordJoiner)\u{2028}\u{2029}over\
+            {the}\(String.wordJoiner)lazy\(String.nonBreakingSpace)dog\n\
+            \u{A}.\u{B},\u{C};\u{D}
+            """
+
+        XCTAssertEqual(original.nonLineBreaking(), expected)
+    }
```
aclima93, line 174: The semantics of this test are a bit misleading 😅 as `\n` is also known as a "line break" but they are expected to be preserved when calling `.nonLineBreaking()` 😵 . I've tried and failed to come up with an alternative method name 😞 especially since upon futher inspection the function documentation clearly states which characters are considered "line breaking" 🤷

aclima93 (approved):

Sources/Extensions/Foundation/String.swift
p4checo: indeed! copy paste fail 😅

Tests/AlicerceTests/Extensions/Foundation/StringTestCase.swift
p4checo: This test was initially designed to test whether newlines were skipped or not, but after some iterations it reached this form where it tests that they're actually preserved. 🙈 I'll rephrase the test name to make it a bit clearer.

The main purpose of the `nonLineBreaking()` helper is to avoid breaking text where it's usually broken when rendered by the system and you don't "control it" - thus we "force" it to via non breaking characters.  It's not to make the string a single "unbreaked" line (which would then require us to convert/strip newlines too). 

As mentioned in a [previous discussion](https://github.com/Mindera/Alicerce/pull/227#discussion_r607202291), automatically converting newlines is not straightforward, so it's safer to preserve them by default. As an escape hatch, we provide the "raw" `replacingOccurrencesOfCharacters` so that users can do whatever they want. The method has been further documented to clearly state that newlines are preserved.

Hope this helped clarify the reasoning. 😄 

Do you have any concrete suggestions?

aclima93: one of the names that came to mind was `reduceUnexpectedLineBreaks()` but I'm the first to admit that it's not a very good name, so I'm perfectly content with your explanation 🙈

p4checo: well, we could add in an extra knob (parameter), which would allow callers to choose what substitution to use for newlines. Something like:

`public func nonLineBreaking(newlineCharacterReplacement: String? = nil) -> String`

Which unlocks the following capabilities depending on the `newlineCharacterReplacement` value:
- `nil`: do nothing - newlines are preserved
- `""`: strip newlines
- `String.nonBreakingSpace`: one line to rule them all
- `"🍌"`: make monkey users happy

I guess that this improves the current API by tying these "loose ends" and giving even more control.

aclima93: Bonus points 💯⭐ for the default value that preserves the current calls 👍

PR #232, opened 2021-04-15
Add single module SPM product 💧

p4checo:
### Checklist

### Motivation and Context

With the latest changes to Package.swift, importing Alicerce via SPM resulted in `import Alicerce` no longer working, as the code is made available in the new submodules (e.g. `AlicerceCore`, `AlicerceLogging`, ...).

This is caused by the fact that SPM targets are the exposed modules, and not the products (which can contain multiple targets). The solution then becomes a matter of creating a new target that includes everything under `Sources`, and a new product that includes just this one target.

However, as SPM performs some checks to ensure path exclusivity (to avoid duplicate symbols), we can't simply create another target that includes everything under `Sources`. One workaround is to trick SPM by creating a symlink to Sources with a different name 👻 [source](https://forums.swift.org/t/spm-shared-targets-files-use-case-whats-the-alternative/38888/4)

That being said, until SPM supports more complex setups, I think we can get away with this. Users will have to be careful to not simultaneously import the single module product with any of the multi module products, as that will surely cause duplicate symbols 💥.

### Description

- Migrate Package.swift to `swift-tools-version:5.3` and fix warnings caused by missing exclude paths in `AlicerceCore` - due to it using `Sources` as `path`, and only `Shared` and `Utils` as sources.

- Create `SingleModuleSources` symlink to `Sources`.

- Create new `AlicerceSingleModule` product and `Alicerce` target in `Package.swift`.

### Why Draft? 🤔 Let's discuss it...

Supporting both single and multi modules simultaneously can be seen as a bit of an anti-pattern, because it's technically possible for users to import products from both single and multi module "groups", effectively causing a duplicate symbol scenario which is what SPM tries to avoid by doing the exclusive path checks.

Nonetheless, I opened this PR as a Draft to start the discussion and see what everyone thinks. For completeness' sake, the main reason I think it might be worth us merging this as-is (vs waiting for this sort of setups to be officially supported by SPM) is because the migration into sub-module-only setup via SPM was unintended 😅. I always thought that the Alicerce product would produce a single module Alicerce, avoiding breaking changes with existing code. Having this single module escape hatch, even if achieved in a bit of a hacky way, provides backwards compatibility and allows for a smoother sub-module adoption.


filipe-lemos (approved):
My opinion is that users of this framework should eventually switch to multiple imports but I agree that's an unintended breaking change and, at least for now, we should keep supporting a single module since it's an option with this hammerito 🔨

diogoeiras (approved):

PR #236, opened 2021-05-11
Unify Release and CI workflows 🤝

p4checo:
### Checklist

### Motivation and Context

Unfortunately, `workflow_run` triggers only capture the branch that triggered the workflow(s) they depend on, which means that we can't use them while filtering for tag pushes (which is what we want for triggering releases). That's why `startsWith(github.ref, 'refs/tags/')` was never `true` when the `Release` workflow ran.

After quite a bit of experimentation and research, I came to the conclusion that we should just keep it simple (KISS and all that 🙈) and just unify our whole CI into a single workflow which easily achieves the setup we want.

Hopefully third time will be the charm to get our releases working via GitHub Actions 🍀

### Description

- Rename `Release` workflow jobs `deploy-github` and `deploy-cocoapods` to `release-github` and `release-cocoapods` respectively.

- Move release workflow jobs to the `CI` Workflow, configured to only run on tag pushes (`X.Y.Z`) and with dependencies on `build-test`, `swiftpm`, `cocoapods` and `carthage` jobs.


.github/workflows/ci.yml
```diff
@@ -111,3 +111,55 @@ jobs:
...
+
+      - name: create release
+        uses: softprops/action-gh-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          body: |
+            # Changes
+
+            - <!-- Insert changes here -->
```
diogoeiras, line 132: Some leftovers?

p4checo: No, this was already in the release workflow job 😄 

It's more for us to already have some kind of format. Going forward we could get fancier and point to a CHANGELOG file or calculate the diff and automate this

filipe-lemos (approved):

diogoeiras (approved):

PR #244, opened 2021-08-03
Integrate Danger in CI

lacerda-eduardo:
### Checklist

### Motivation and Context
Install Danger-Swift (https://danger.systems/swift/) to automate common code reviews rules. Danger will write comments on PR as 'github-actions' bot.
Danger is installed just to run on CI, it not a project dependency.

### Description
- Add danger on brew file to be installed at CI.
- Create Dangerfile.swift, at this file is configured all rules that Danger will check on PRs. For now we just have SwiftLint rules and a List of all changed files. 
- Change ci.yml to add Danger step. It runs 417-72KI/danger-swiftlint@v3 GitHub Action.


.github/workflows/ci.yml
```diff
@@ -112,6 +112,16 @@ jobs:
       - name: carthage build
         run: ./script/carthage.sh build --cache-builds --no-skip-current
 
+  Danger:
+    name: Danger Swift
+    runs-on: ubuntu-latest
```
diogoeiras, line 117: Maybe lock to the specific version as we do for Mac? `ubuntu-20.04`

Dangerfile.swift
```diff
@@ -0,0 +1,12 @@
...
+import Danger
+
+let danger = Danger()
+
+// Swiftlint style check, with comment on PR
+SwiftLint.lint(.all(directory: nil), inline: true)
+
+// List of all edited files
+let editedFiles = danger.git.modifiedFiles + danger.git.createdFiles
+message("These files have changed: \(editedFiles.joined(separator: ", "))")
```
diogoeiras, line 12: Do we need to list files changed? What do we gain with this? Also, we're missing `danger.git.deletedFiles`

diogoeiras (approved):

Dangerfile.swift
lacerda-eduardo: @diogoeiras I removed this rules (was using for test). Keep only SwiftLint until we have more rules to this specific project.

.github/workflows/ci.yml
```diff
@@ -112,6 +112,16 @@ jobs:
       - name: carthage build
         run: ./script/carthage.sh build --cache-builds --no-skip-current
 
+  Danger:
```
p4checo, line 115: I think this could live in a separate action instead, like `danger.yml` which only runs on PRs

Brewfile
```diff
@@ -1 +1,3 @@
 brew 'Swiftlint'
+tap "danger/tap"
+brew "danger-swift"
```
p4checo, line 3: are these changes still needed, given Danger is provided by the `417-72KI/danger-swiftlint@v3` action?

Dangerfile.swift
```diff
@@ -0,0 +1,8 @@
+// Copyright © 2021 Mindera. All rights reserved.
```
p4checo, line 1: I think we don't need this copyright notice

p4checo (changes requested):
Nice work! Still think we can polish it a bit though 🤓

Brewfile
lacerda-eduardo: @p4checo do you think we should keep Swiflint? It is now integrated with Danger, maybe we could remove it too.

p4checo: I think we can. 

It was added here during the initial SwiftLint integration in the project, but funnily enough it was never invoked via CI 😅

We also run SwiftLint locally in the project, which is done via a SwiftLint target that contains a shell script to run `swiftlint` and print a warning if it isn't installed. In hindsight, this script should trigger an error instead of just printing... 🤔

Dangerfile.swift
```diff
@@ -0,0 +1,6 @@
+import Danger
+
+let danger = Danger()
+
+// Swiftlint style check, with comment on PR
+SwiftLint.lint(.all(directory: nil), inline: true)
```
p4checo, line 6: perhaps add a trailing newline?

.github/workflows/danger-ci.yml
```diff
@@ -0,0 +1,19 @@
+name: Danger-CI
```
p4checo, line 1: perhaps just `Danger` 🤓 💅🏼

Dangerfile.swift
```diff
@@ -0,0 +1,7 @@
...
+
```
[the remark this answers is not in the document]
p4checo: I don't get why SwiftLint is complaining here 🤔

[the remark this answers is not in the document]
lacerda-eduardo: Me neither.

p4checo (approved):
🚀

filipe-lemos (approved):

PR #246, opened 2021-09-01
Make `TrieRouter` `register`/`unregister` receive `URL`s 📒

p4checo:
### Checklist

### Motivation and Context

When we updated `Router` and `RouteHandler` to have an additional `R: Routable` associated type, we also migrated all public `TrieRouter` APIs to receive an `R` instance, replacing `URL`.

However, the main purpose of the new `R: Routable` is to allow having more data when effectively _being routed_ (i.e. in `route()`), and not on all other "events", like registering/unregistering routes.

So, if we for instance define `R` as being some sort of HTTP request type (which makes total sense to be routed), it doesn't make much sense to use that same type to register/unregister routes, because it forces us to create "dummy" requests only for those purposes.

With that in mind, `TrieRouter` `register()` and `unregister()` now expect (annotated) URLs like before, and only receive an `R` instance in `route()`.

### Description

- Update `TrieRouter` `register()` and `unregister()` receive a `URL` route.

- Update some `TrieRouter` public API documentation.

### DIFF
--- a/Sources/DeepLinking/Route+TrieRouter.swift
+++ b/Sources/DeepLinking/Route+TrieRouter.swift
@@ -78,15 +78,22 @@ extension Route {
         /// Creates an instance of a trie router.
         public init() {}
 
-        /// Registers a new route in the router with the given handler.
+        /// Registers a new *annotated URL* route in the router with the given handler.
+        ///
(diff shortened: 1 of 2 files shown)


p4checo:
Thanks for the _supersonic_ review @filipe-lemos! 🙇🏼

PR #249, opened 2022-01-17
Revise `Box`es 📦

p4checo:
### Checklist

### Motivation and Context

We have had two `Box` variants for a very long time: `Box` and `VarBox`, containing an immutable and mutable value respectively.

However, from our experience the immutable variant is not used often if at all, which means its immutability guarantee is not something that people want or need.

Furthermore, following the latest language developments a `Box` is a very good candidate for a property wrapper, and more so now that they can be defined as local properties.

### Description

- Deprecate `VarBox` in favor of `Box`, making it wrap a mutable value.

- Make `Box` a `@propertyWrapper`, and `@dynamicMemberLookup`. The latter is useful if we use `Box<T>` as a regular type (i.e. not a PW).


p4checo:
It seems CI is breaking because it's using `macOS-11` image, which currently still compiles with Xcode 12.5. That will apparently [change on January 24th](https://github.com/actions/virtual-environments/issues/4800), so we can wait a couple of days to avoid messing with CI to override the active Xcode version.

An official `macOS-12` image should pop up eventually as well 🙏🏼

Sources/Utils/Box.swift
```diff
@@ -1,33 +1,39 @@
...
-public final class VarBox<T> {
+extension Box {
 
-    /// The encapsulated value.
-    public var value: T
+    /// The wrapped value (compact).
+    public var value: T {
+        get { wrappedValue }
+        set { wrappedValue = newValue }
+    }
```
aclima93, line 28: it appears that internally only Network+URLSessionNetworkStack.swift used Box, are we keeping this `value` so that it doesn't break existing codebases that rely on alicerce? if so, won't it also be confusing exposing `wrappedValue` as well? maybe `wrappedValue` could be private, keeping older codebases compliant but not obfuscating the internals of what it actually does (wrapping a value with a propertyWrapper).

Or maybe i'm just over-analyzing all of this 😅

p4checo: > are we keeping this value so that it doesn't break existing codebases that rely on alicerce? 
Yes, the main motivation is to make the transition smoother. Also, `Box<T>` can now be used both as a PW and as a regular type, so when used as a regular type I thought it would be a bit more ergonomic to have a `value` property, as well as an `init` without the parameter name.

> if so, won't it also be confusing exposing `wrappedValue` as well?
I don't think this will be confusing, as `wrappedValue` is well known to be tightly coupled to PWs. Also, users can always look into the code to see what's going on, right? We even have documentation (e.g. `(compact)`) in place. 🤓 

> maybe `wrappedValue` could be private, keeping older codebases compliant but not obfuscating the internals of what it actually does (wrapping a value with a propertyWrapper).
Sadly this doesn't work, because `wrappedValue` is a requirement of PWs, and must have the same access level as its enclosing PW. A similar thing happens with the `init(wrappedValue: T)`

Screen Shot 2022-01-18 at 09 22 09

Hope this answers your questions 😄

aclima93: it does, thanks! 🙇

aclima93 (approved):

PR #254, opened 2022-07-25
Add `aspectRatio` constraint helper

filipe-lemos:
### Checklist

### Motivation and Context
Some layouts are more easily achieved by defining a specific aspect
ratio for a view which, when combined with constraints to establish the
width or height of the view, allows Auto Layout to infer the size in
the other dimension.

### Description
Add `aspectRatio` constraint helper through a new `AspectRatioConstrainableProxy`. This allows the creation of relative constraints between the width and height dimensions of a view.


Sources/AutoLayout/ConstrainableProxy.swift
```diff
@@ -844,7 +867,8 @@ public extension BaselineConstrainableProxy {
     }
 }
 
-public protocol PositionConstrainableProxy: EdgesConstrainableProxy, SizeConstrainableProxy, CenterConstrainableProxy {}
+public protocol PositionConstrainableProxy:
+    EdgesConstrainableProxy, SizeConstrainableProxy, AspectRatioConstrainableProxy, CenterConstrainableProxy {}
```
p4checo, line 871: shouldn't `AspectRatioConstrainableProxy` be a part of `SizeConstrainableProxy` instead, since it effectively only affects size? (and not position)

One could argue the `aspectRatio` could even be just one more operator provided by `SizeConstrainableProxy`, instead of creating `AspectRatioConstrainableProxy` just for it 🤔

filipe-lemos: I had it originally on `SizeConstrainableProxy` but thought it might seem out of place. Will put it back there, thanks!

p4checo (approved):
🚀

PR #270, opened 2024-05-17
Enable fork PRs CI to run codecov

p4checo:
### Checklist

### Motivation and Context

For security reasons, fork PRs don't have access to secrets if we use `pull_request` in GH Actions CI spec, only if we use `pull_request_target`, which has its own security implications. See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Codecov status reporting in CI requires a token, so for fork PRs to be able to do so we migrated to `pull_request_target` with the caveat that MRs have to be labelled and have the `run ci` label applied, which can only be done by someone with triage access to the repo.

This should give us a good compromise in terms of security.

### Note ⚠️ 

CI isn't running with these changes, because `pull_request_target` runs the workflow from the **target branch,** so only once this gets merged to `master` will the workflow actually run as "expected".

### Description

Update `ci.yml` specification to allow fork PRs CI to run codecov, but only when labelled with the `run ci` label.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - master
     tags:
       - '[0-9]+\.[0-9]+\.[0-9]+'
-  pull_request:
+  pull_request_target: # forks don't have access to secrets if we use `pull_request`, which is required for codecov
     branches:
       - master
 
@@ -24,6 +24,10 @@ jobs:
   env-details:
     name: Environment details
     runs-on: macos-14
+    if: |
+      github.event_name == 'push' ||
+      !github.event.pull_request.head.repo.fork ||
+      (github.event.pull_request.head.repo.fork && contains(github.event.pull_request.labels.*.name, 'run ci'))
     steps:
       - name: xcode version
         run: xcodebuild -version -sdk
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ on:
       - master
     tags:
       - '[0-9]+\.[0-9]+\.[0-9]+'
-  pull_request:
+  pull_request_target: # forks don't have access to secrets if we use `pull_request`, which is required for codecov
     branches:
       - master
```
pedroseruca, line 12: I was wondering if we could / should centralise this condition since is used multiple times?

inside `env:` we could add a new line:
```yml
RUN_CI_CONDITION: ${{ github.event_name == 'push' || !github.event.pull_request.head.repo.fork || (github.event.pull_request.head.repo.fork && contains(github.event.pull_request.labels.*.name, 'run ci')) }}
```

and then use it below:

```yml
if: env.RUN_CI_CONDITION
```

does this makes sense?

p4checo: Thanks for your suggestion! 

I agree that we should try to avoid this duplication, and I think this could work but we would probably need to evaluate the env var via `if: {{ env.RUN_CI_CONDITION }}` so that it would coerce to an actual Boolean value.

However, I am not sure if this approach is a good idea because env vars can be overridden (e.g. by writing to `$GITHUB_ENV`), which might be used to circumvent the _protection_ that the `if` condition is supposed to do.

I would be more comfortable if there was some sort of function/macro that we could define which would be evaluated in each `if`, preventing the actual flag value from being tampered with.

Do you agree?

pedroseruca: Seems good to me 👍🏻

